PR #65, opened 2025-05-21
(SP: 3) Implement Locale-based Routing and Middleware

### LINKED ISSUE #28: (SP: 3) Implement Locale-based Routing and Middleware
Implement a routing system that includes locale prefixes in all URLs (e.g., /uk/about-us, /en/about-us) and build a middleware layer that ensures correct redirection and persistence of the user's language preference throughout their browsing experience. This will ensure consistency in multilingual navigation and a smooth user experience.

**Implementation Requirements:**

**1. Locale-Prefixed Routing:**

     - All public-facing URLs must include a locale prefix (/uk, /en, etc.).

     - Route paths should follow this format: /:locale/<page>.

**2. Locale Switch Handling:**

     -  When the user changes the language:

          - The locale prefix in the URL must be updated.

          - The user should stay on the same logical page (e.g., /uk/about-us → /en/about-us).

**3. Locale Persistence via Cookies:**   
 
     - The selected language must be stored in cookies (e.g., locale=uk).

     - Use appropriate cookie settings:

          - Long-lived expiration (e.g., 1 year).

**4. Middleware for Default Redirects:**

     - Middleware should automatically:

          - Detect when a user accesses a non-prefixed route (e.g., /about-us).

          - Redirect them to the same route with the appropriate locale prefix:

               - From /about-us → /uk/about-us (based on cookie or fallback to default).

          - Middleware must run early in the request lifecycle (server-side or edge depending on framework).

Linked user story: https://github.com/Liatoshynsky-Foundation/lf-client/issues/10

Iarynovskyi:
## Summary of issue

Implement locale-based routing via middleware in a Next.js application to support multilingual navigation with persistent language preferences.

## Summary of change
- Added `middleware.ts` at the root level (as required by Next.js v13+),  it should be in a root with `next.config.ts`
- Added base constants to `constants/index.ts`
- Added utils to work with locales and cookies
- Tried to add aliases `~/...` to `tsconfig.json` and `jest.config.ts`
- Added unit tests 

## Result on site

https://github.com/user-attachments/assets/378ad8f0-0453-49bf-9e9e-8296c9f55aa7

## Result of tests

Знімок екрана 2025-05-21 о 18 08 23


app/local-based-url-middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+        }
+        return NextResponse.next();
+    }
+
+    const locale = determineLocale(localeFromCookie || DEFAULT_LOCALE);
+    const redirectUrl = new URL(`/${locale}${pathname}${search}`, request.url);
+    const response = NextResponse.redirect(redirectUrl);
+    setCookie(response, locale, LANGUAGE_COOKIES, ONE_MONTH_IN_SECONDS);
+
+    return response;
```
bandvov: this can be done simplier.
```javascript
    setCookie(response, locale, LANGUAGE_COOKIES, ONE_MONTH_IN_SECONDS);
return NextResponse.redirect(redirectUrl);
```

app/local-based-url-middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    if (isStaticOrApiPath(pathname)) {
+        return NextResponse.next();
+    }
+
+    const localeInPath = extractLocaleFromPath(pathname);
+    const localeFromCookie = getCookie(LANGUAGE_COOKIES, request);
+
+    if (localeInPath) {
+        if (localeFromCookie !== localeInPath) {
```
bandvov: row 30 is redundand check

app/local-based-url-middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+    extractLocaleFromPath,
+    determineLocale
+} from 'lib/i18n/utils';
+import {
+    getCookie,
+    setCookie,
+} from 'lib/utils/cookies';
+import { isStaticOrApiPath } from 'lib/utils/is-static-or-api-path';
+
+export const localBasedUrlMiddleware  = (request: NextRequest) => {
```
bandvov: should it have name `localeBasedUrlMiddleware`? same question about filename

app/constants/index.ts
```diff
@@ -0,0 +1,4 @@
+export const SUPPORTED_LOCALES = ['ua', 'en'];
+export const DEFAULT_LOCALE = 'ua';
```
bandvov: I think ukrainian locale should be "uk", not "ua". Please double-check naming for locale used by i18n.

app/lib/utils/is-static-or-api-path.ts
```diff
@@ -0,0 +1,9 @@
+export const isStaticOrApiPath = (
+    pathname: string
+): boolean => {
+    const ignoredPrefixes = [
+        '/api',
+        '/.next/static'
```
bandvov: should it be `'/_next/static`?

app/local-based-url-middleware.ts
Iarynovskyi: I need here `const response = ...` for `setCookie(response, ...)` in a next line

app/local-based-url-middleware.ts
Iarynovskyi: I will think about a better way to rename it, thanks

app/constants/index.ts
Iarynovskyi: i saw 'ua' in figma and something stuck in my head
 it is my fault, thanks

app/lib/utils/is-static-or-api-path.ts
Iarynovskyi: you are right, thanks
the `/.next` dir confused me

app/local-based-url-middleware.ts
bandvov: oh. yes, you are right!

app/local-based-url-middleware.ts
bandvov: name is good. probably there is typo(aka should be locale not local)?

app/constants/index.ts
bandvov: No worries

app/local-based-url-middleware.ts
Iarynovskyi: oh, now 'locale' written correctly
'e' escaped me, but i catch it now

app/url-locale-middleware.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+    getCookie,
+    setCookie,
+} from 'lib/utils/cookies';
+import { isStaticOrApiPath } from 'lib/utils/is-static-or-api-path';
+
+export const urlLocaleMiddleware  = (request: NextRequest) => {
+    const { pathname, search } = request.nextUrl;
+
+    if (isStaticOrApiPath(pathname)) {
```
ShadowOfTheSpace: You don't need that check at all, since you have middleware matcher that excludes api and `_next/..` paths. Next.js middleware won't be executed on those paths.

app/lib/i18n/utils.test.ts
ShadowOfTheSpace, line 1: You have typos in titles of tests, fix those please

app/lib/i18n/utils.ts
```diff
@@ -0,0 +1,21 @@
+import {
+    DEFAULT_LOCALE,
+    SUPPORTED_LOCALES
+} from '~/constants';
+
+export const extractLocaleFromPath = (
+    pathname: string
+): string | null => {
+    const segments = pathname.split('/');
```
ShadowOfTheSpace: You don't need to split all of the URL into parts; you need only the second part. So `split` has the second param for that - stop after a second part.
Since middleware should be as fast as possible, I think it's a nice little improvement in performance.
```suggestion
    const segments = pathname.split('/', 2);
```

app/lib/utils/cookies.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export const setCookie = (
+    response: CookieSetter,
+    cookieValue: string,
+    cookie: string,
+    maxAge: number
+): void => {
+    response.cookies.set(cookie, cookieValue, {
+        maxAge: maxAge,
```
ShadowOfTheSpace:
```suggestion
        maxAge,
```

middleware.ts
```diff
@@ -0,0 +1,12 @@
+import type { NextRequest } from 'next/server';
+import { urlLocaleMiddleware } from '~/url-locale-middleware';
+
+export const middleware = (request: NextRequest) => {
+    return urlLocaleMiddleware(request);
+}
+
+export const config = {
+    matcher: [
+        '/((?!api|_next/static).*)',
```
ShadowOfTheSpace: I also suggest you include that in the matcher. Next will replace all of the image URLs with optimized ones (starting from _next/image). Also, you definitely want to receive your favicon, so disable the language check on that one too.
```suggestion
        '/((?!api|_next/static|_next/image|favicon.ico).*)',
```

app/lib/i18n/utils.ts
Iarynovskyi: I agree with you, thanks

Mike-Popovych (approved):

danikua (approved):

VKormylo (approved):

yur4uwe (approved):

ShadowOfTheSpace (approved):

PR #79, opened 2025-05-24
(SP: 3) Create Checkbox element aligned with design system 

### LINKED ISSUE #67: (SP: 3) Create Checkbox element aligned with design system
Create a Checkbox element for universal usage through the design system based on designs from Figma

Requirements:

should follow the design based on the current mock-ups in Figma
should be usable universally across the app
Props:

states (default, hover focused, diasble, selected)

Image

luvthenika:
## Summary of issue

- [x] implemented reusable checkbox component
- [x] styled checkbox component according to the design system
- [x] covered checkbox with tests
hovered: 
image
focused:
image
disabled:
image
selected:
image

### DIFF
--- a/app/components/Checkbox/Checkbox.spec.tsx
+++ b/app/components/Checkbox/Checkbox.spec.tsx
@@ -1,0 +1,35 @@
+import { render, screen } from '@testing-library/react';
+import CustomCheckbox from './Checkbox';
+import userEvent from '@testing-library/user-event';
+
+describe('CustomCheckbox', () => {
+    const label = 'Label';
+    it('should display checkbox', async () => {
+        render(<CustomCheckbox label={label} />);
+        const checkbox = screen.getByRole('checkbox');
+        expect(checkbox).toBeInTheDocument();
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.getByRole('checkbox')).toBeChecked();
+    });
+    it('should check and uncheck the checkbox', async () => {
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.getByRole('checkbox')).toBeChecked();
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+
+    });
+    it('should not toggle disabled checkbox', async () => {
+        const disabled = true;
+        render(<CustomCheckbox disabled={disabled} label={label} />);
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+    });
+    it('should display checked checkbox by default', async () => {
+        const disabled = true;
+        render(<CustomCheckbox disabled={disabled} label={label} />);
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+    });
+
+});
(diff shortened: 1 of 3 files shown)


app/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+    it('should display checkbox', async () => {
+        render(<CustomCheckbox label={label} />);
+        const checkbox = screen.getByRole('checkbox');
+        expect(checkbox).toBeInTheDocument();
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.getByRole('checkbox')).toBeChecked();
+    });
+    it('should check and uncheck the checkbox', async () => {
+        await userEvent.click(screen.getByRole('checkbox'));
```
Iarynovskyi, line 15: You forgot to add CustomCheckbox render here

app/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+        expect(screen.getByRole('checkbox')).toBeChecked();
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+
+    });
+    it('should not toggle disabled checkbox', async () => {
+        const disabled = true;
+        render(<CustomCheckbox disabled={disabled} label={label} />);
```
Iarynovskyi:
```suggestion
        render(<CustomCheckbox disabled label={label} />);
```
I am not sure but it seems like it might work without `const disabled = true;` 
And if it is correct so do like this in next test case too

app/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,30 @@
+'use client';
+import { Checkbox } from '@mui/material';
+import FormGroup from '@mui/material/FormGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import { CheckboxProps } from '@mui/material/Checkbox';
+import { FormControlLabelProps } from '@mui/material/FormControlLabel';
```
Iarynovskyi: Double import, you can combine they into one

Iarynovskyi (changes requested):

app/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+    });
+    it('should display checked checkbox by default', async () => {
+        const disabled = true;
+        render(<CustomCheckbox disabled={disabled} label={label} />);
+        await userEvent.click(screen.getByRole('checkbox'));
+        expect(screen.queryByRole('checkbox')).not.toBeChecked();
+    });
```
oleg191006, line 28: It seems that these 2 tests repeat the same logic of checking for disabled checkbox. If the last test was to check that the checkbox is displayed as checked by default, then you need to explicitly pass the defaultChecked prop, otherwise it tests the same as the previous one

Iarynovskyi (approved):

bandvov:
I think it would be better to have such props:
```
{ disabled, checked, onChange, size, label }
```
 In this case we will have conntrollable component, with possibility to store state outside component.
Any concerns?

IrynaKhylchuk (approved):

danikua (approved):

PR #83, opened 2025-05-24
(SP: 2) Integrate next-intl and Configure Localization

### LINKED ISSUE #34: (SP: 2) Handle Missing Translation Pages Gracefully
Implement a fallback page that informs users when the requested content is not yet available in the selected language. This ensures a smooth and user-friendly experience even when translation coverage is incomplete.

**Translation Check:**
- Implement logic that determines whether the translation/content for a specific page and locale exists.

**Fallback Page Display:**
- If the translation is unavailable:
       Render a dedicated "Content Unavailable" page or component.

- Text must include:   "Сторінка поки що недоступна цією мовою"

- Provide a button:     
Text: "На головну"
Action: Navigates to the localized homepage (/uk or /en based on current locale).

Design mockup: --

### LINKED ISSUE #57: (SP: 2) Integrate i18next and Configure Localization
- Initialize and configure` react-i18next` with two languages (UA, EN).
- Set Ukrainian (UA) as the default language.
- Ensure the selected language is saved in localStorage, so the preference persists across page navigation, refreshes, and future visits in the same browser.
- Update page content dynamically according to the selected language without page reload.

Iarynovskyi:
## Summary of change

- Implemated `next-intl` for localization and refactored middleware in project
- All work with SSR without `'use client'`
- 'UA' as default language
- Added translations for some pages
- Refactor some tests, components and pages
- Added test NotFound page for cases when we have not some translations for this page
This NotFound page i tried to do dynamic for not found language, page and more, but in NextJs it is harder than i think before to implement this, so it will remain as a new issue
 **But we need in future to change this page after we get the design**

## Result

https://github.com/user-attachments/assets/307fde71-cc50-4da2-9cb5-e4dbbb48e2bd

## No ESLint warnings or errors and all tests passed

Знімок екрана 2025-05-24 о 15 08 18


app/lib/i18n/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        .use(initReactI18next)
+        .use(resourcesToBackend((language: string, namespace: string) =>
+            import(`~/constants/locales/${language}/${namespace}.json`)
+        ))
+        .init(getOptions(lang, ns));
+    return i18nInstance;
+};
+
+export async function getTranslations(lang: string, ns?: string | string[]) {
+    const i18nextInstance = await initI18next(lang, ns || getOptions().DEFAULT_NS);
```
bandvov: consider using some sort of caching here(Map for example) this should help to prevent unnecessary inits of i18n instance.

Iarynovskyi: thanks for advice, now i have added `Map` caching here

app/[lang]/layout.tsx
```diff
@@ -20,21 +23,27 @@ export const metadata: Metadata = {
...
+    <html lang={lang}>
       <body className={`${geistSans.variable} ${geistMono.variable}`}>
         <Container
           maxWidth="md"
           sx={{ border: '1px solid #ccc', padding: '20px', minWidth: '100%' }}
         >
           <Header />
           {children}
-          <Footer />
+          <Footer lang={lang}/>
```
Comment (ShadowOfTheSpace) [inline app/[lang]/layout.tsx]: Do we always need to have prop drilling for passing the current locale with this i18n lib?

app/lib/i18n/utils.ts
```diff
@@ -18,4 +18,14 @@ export const determineLocale = (localeFromCookie: string): string => {
...
+};
+
+export const isPublicLocalPath = (
+    pathname: string
+): boolean => {
+    const locale = extractLocaleFromPath(pathname);
+    const pathWithoutLocale = locale ? pathname.replace(`/${locale}`, '') || '/' : pathname;
+
+    return /\.(svg|png|jpg|ico)$/.test(pathWithoutLocale);
+
```
ShadowOfTheSpace: Adding `.*\\..*` to the `matcher` can help to skip middleware invication on public (static) files.

app/[lang]/layout.tsx
Comment (Iarynovskyi) [reply app/[lang]/layout.tsx]: Only in Header and Footer here because they not from [lang] dir, but from components dir
Another ones get it from [lang] dir

app/lib/i18n/utils.ts
Iarynovskyi: Thanks. Now i understand it!
But i will make this change later after merging this PR #86 where @yur4uwe  fixed it

app/[lang]/layout.tsx
Comment (ShadowOfTheSpace) [reply app/[lang]/layout.tsx]: Yup, got that, but my question is about the process of getting the current locale. 

For example, with `next-intl` I can get a locale in every server component by looking on the specific header with locale (it works like that under the hood). Since headers access makes your page dynamic, there is a wrapper that are usinig React.cache with no direct access to the headers. You set the locale ones in every page - and you will have current locale in every server static component.

I know that you are not using this lib, but [it](https://next-intl.dev/docs/getting-started/app-router/with-i18n-routing#add-setrequestlocale-to-all-relevant-layouts-and-pages) might be helpful.

It's worrying me because I see here the React anti-pattern (props drilling) and it will be increasing since you will have more server components for each page.

Comment (Iarynovskyi) [reply app/[lang]/layout.tsx]: Thanks, I will look into this. 
Because I just tried to add retrieving the locale from the cookie and the tests became very complicated.
And I do not really want to rewrite everything for another library, but if I do not find a way out, I will have to do it.

Comment (Iarynovskyi) [reply app/[lang]/layout.tsx]: This library turned out to be much better so I refactored everything to it after discussing it on todays meeting 
Now it works without anti-patterns

Comment (ShadowOfTheSpace) [reply app/[lang]/layout.tsx]: It's very nice that you find out this one much better, hope you like it!

app/[lang]/not-found.tsx
```diff
@@ -0,0 +1,16 @@
+import Link from 'next/link';
+
+export default async function CustomNotFoundPage() {
+  return<>
+    <section style={{ marginBottom: '20px' }}>
+            Сторінка ще недоступна цією мовою
+    </section>
+    <Link href={'/'} passHref style={{
```
Comment (ShadowOfTheSpace) [inline app/[lang]/not-found.tsx]:
```suggestion
    <Link href='/' passHref style={{
```

middleware.ts
```diff
@@ -1,12 +1,10 @@
...
-export const middleware = (request: NextRequest) => {
-  return urlLocaleMiddleware(request);
-};
+export default createMiddleware(routing);
 
 export const config = {
   matcher: [
     '/((?!api|_next/static|_next/image|favicon.ico|.*\\.[^/]+$).*)'
-  ],
+  ]
```
ShadowOfTheSpace: Since you have a single matcher, probably you don't need an array here

i18n/request.ts
ShadowOfTheSpace, line 1: Also suggest you add something like this in one of `next-intl` configuration files:
```ts
import { type useTranslations } from "next-intl";

import { type IsEqual } from "~/libs/types/types";

import { type routing } from "./config";
import { type default as enMessages } from "./messages/en.json";
import { type default as ukMessages } from "./messages/uk.json";

declare module "next-intl" {
	interface AppConfig {
		Locale: (typeof routing.locales)[number];
		Messages: IsEqual<typeof enMessages, typeof ukMessages> extends true
			? typeof ukMessages
			: never;
	}

	type TranslationKey = Parameters<ReturnType<typeof useTranslations>>[0];
}
```

It will add an autocomplition to `useTranslation` and `getTranslation` calls.
Also, you can skip `IsEqual` part and assign one of the jsons as a type (`IsEqual` part is a custom logic for finding out the you don't have equals translation files - it will disable an autocomplition for translations, and it will be sign that translation is missing).

`TranslationKey` type is needed when you pass a variable (not a constant string) to the `t` function. For example: `t(errorFromApi)` (when `errorFromApi` is a string, not a "some.translation.key" constant). In this situation to not have ts error, you have to cast your error like this: `t(errorFromApi as TranslationKey)`.

This code is totally optional and also requires some modification according to your project, but it can help devs with translation usage.

image

i18n/routing.ts
```diff
@@ -0,0 +1,7 @@
+import {defineRouting} from 'next-intl/routing';
+
+export const routing = defineRouting({
+  locales: ['en', 'uk'],
+  defaultLocale: 'uk',
+  localeDetection: false
```
ShadowOfTheSpace, line 6: Why did you disable this?

i18n/request.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default getRequestConfig(async ({requestLocale}) => {
+  const requested = await requestLocale;
+  const locale = hasLocale(routing.locales, requested)
+    ? requested
+    : routing.defaultLocale;
+
+  return {
+    locale,
+    messages: (await import(`~/constants/messages/${locale}.json`)).default,
```
ShadowOfTheSpace: I suggest you to not put messages in the `constant` folder. Put this folder inside `i18n` folder just to be consistant.

app/[lang]/collaboration/page.tsx
```diff
@@ -1,5 +1,6 @@
-import React, { ReactElement } from 'react';
+import { getTranslations } from 'next-intl/server';
 
-export default function Collaboration(): ReactElement {
-  return <div>Collaboration</div>;
+export default async function CollaborationPage() {
+  const t = await getTranslations('collaborations');
```
Comment (ShadowOfTheSpace) [inline app/[lang]/collaboration/page.tsx:12]: In all static pages (not dynamic ones) you have to call `setRequestLocale(locale);` with a `locale` from URL parameters. Otherwise, you won't have static abilities even on real static pages (with no requests inside).

app/[lang]/not-found.tsx
```diff
@@ -0,0 +1,16 @@
+import Link from 'next/link';
+
+export default async function CustomNotFoundPage() {
```
Comment (ShadowOfTheSpace) [inline app/[lang]/not-found.tsx:4]: This `notFound` page won't be called when you visit a non-existing page (`/uk/not-existing-page`). The default NextJS not found page will be shown instead.
To fix that - you have to create a `catch-all` route in the `[lang]` folder with just a `notFound` function call.
It's the only one (as I know) way to have a translated NotFound page.

```ts
import { notFound } from "next/navigation";

const UnknownRoutePage: React.FC = () => {
	notFound();
};

export default UnknownRoutePage;
```

ShadowOfTheSpace (commented):
Also, forgot to add. 
Do you guys have a configured prettier for the project? In a lot of PRs, I see unformatted files. 
**It's very important to have consistent formatting with all devs on the project**, so, please, add a pre-commit hook to format your code before pushing it.
Also, if needed, you can always check the `s2s` project with already configured formatting checks (local and gh-actions) to always have formatted files in the project).

Iarynovskyi:
> Also, forgot to add. Do you guys have a configured prettier for the project? In a lot of PRs, I see unformatted files. **It's very important to have consistent formatting with all devs on the project**, so, please, add a pre-commit hook to format your code before pushing it. Also, if needed, you can always check the `s2s` project with already configured formatting checks (local and gh-actions) to always have formatted files in the project).

I now starting work on this task

i18n/routing.ts
Iarynovskyi: Because without this it takes the language from the browser by default and for example mine is not UK

app/[lang]/not-found.tsx
Comment (Iarynovskyi) [reply app/[lang]/not-found.tsx:4]: You are genius! Thanks!

app/[lang]/collaboration/page.tsx
Comment (Iarynovskyi) [reply app/[lang]/collaboration/page.tsx:12]: I tried to fix it but something tells me that I did not do what you meant

Comment (ShadowOfTheSpace) [reply app/[lang]/collaboration/page.tsx:12]: I'll check that, no worries)

i18n/routing.ts
ShadowOfTheSpace: Got that. I haven't used middleware from `next-intl`, I created my own one with custom logic. Firstly, I check the language from the `URL`, after that from cookies, and lastly - from the `accept-language` header. If there are no accepted languages - use the default one.
I thought that this lib has the same logic, isn't it?

I'll check that today or tomorrow and will tell you, maybe there is a workaround for this.

i18n/request.ts
ShadowOfTheSpace: Have you tried that? Does it work correctly with the current lib version?

Iarynovskyi: Yes, i tried it. And it works as you show in screenshot
And if i write something else it will emphasize it like error

app/[lang]/[not-found]/page.tsx
Comment (ShadowOfTheSpace) [inline app/[lang]/[not-found]/page.tsx:1]: Probably you didn't understand it correcly, under `catch-all` route I meant [this](https://nextjs.org/docs/app/api-reference/file-conventions/dynamic-routes#catch-all-segments). 
`/app/[lang]/[...unknown-route]/page.tsx`

app/[lang]/layout.tsx
```diff
@@ -20,23 +22,31 @@ export const metadata: Metadata = {
   description: 'Generated by create next app',
 };
 
-export default function RootLayout({
+interface RootLayoutParams {
+  readonly children: ReactNode;
+  readonly params: Promise<{ readonly lang: 'en' | 'uk' }>
```
Comment (ShadowOfTheSpace) [inline app/[lang]/layout.tsx]: `import { Locale } from "next-intl"`
```suggestion
  readonly params: Promise<{ readonly lang: Locale }>
```

app/types/types/lang.type.ts
```diff
@@ -0,0 +1,3 @@
+export interface LangType {
```
ShadowOfTheSpace: It's not `LangType` it's a type of url params
```suggestion
export interface ParamsWithLanguage{
```

app/types/types/lang.type.ts
```diff
@@ -0,0 +1,3 @@
...
+    readonly params: Promise<{ readonly lang: 'en' | 'uk' }>
```
ShadowOfTheSpace: Same about `Locale` usage

i18n/next-intl.d.ts
```diff
@@ -0,0 +1,20 @@
+import { type useTranslations } from 'next-intl';
+
+import { type routing } from './routing';
+import { type default as enMessages } from './messages/en.json';
+import { type default as ukMessages } from './messages/uk.json';
+
+type IsEqual<T, U> =
+    (<G>() => G extends T ? 1 : 2) extends
+        (<G>() => G extends U ? 1 : 2) ? true : false;
```
ShadowOfTheSpace: I'd change `1` and `2` to `true` and `false` to be more explicit

```ts
type IsEqual<A, B> =
	(<T>() => T extends A ? true : false) extends <T>() => T extends B
		? true
		: false
		? true
		: false;

```

i18n/next-intl.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+declare module 'next-intl' {
+    interface AppConfig {
+        Locale: (typeof routing.locales)[number];
+        Messages: IsEqual<typeof enMessages, typeof ukMessages> extends true
```
ShadowOfTheSpace, line 15: If you would have problems with its performance (long time waiting for autocomplete or smth like that), just remove comparing (probably it takes a lot of time to compare those types).

i18n/request.ts
ShadowOfTheSpace: Cool, very nice!

app/[lang]/collaboration/page.tsx
Comment (ShadowOfTheSpace) [reply app/[lang]/collaboration/page.tsx:12]: Yeah, you did what I meant. But don't forget to write `setRequestLocale(lang)` in **every** page (`page.tsx` file) in the future.
Check [this](https://github.com/amannn/next-intl/blob/main/packages/next-intl/src/server/react-server/RequestLocale.tsx) and [this](https://github.com/amannn/next-intl/blob/main/packages/next-intl/src/server/react-server/RequestLocaleCache.tsx) parts of `next-intl` source code, probably it will help you to understand why you gonna need that in the future.
Also, [check this part of docs](https://github.com/amannn/next-intl/blob/main/packages/next-intl/src/server/react-server/RequestLocaleCache.tsx).

i18n/messages/uk.json
```diff
@@ -0,0 +1,20 @@
+{
+  "collaborations": {
+    "meta_data": "Співпраця Лятошинського",
+    "text": "Співпраця"
+  },
+  "footer": {
+    "copyright": "© 2025 Liotoshynsky Foundation. Всі права захищені.",
+    "link_privacy": "Політика конфіденційності",
```
ShadowOfTheSpace: Just my preference, but I think the camel case will be nicer here (it will look like the default variable names in the project).

i18n/next-intl.d.ts
Iarynovskyi: Personally me did not have any problems, but if anyone has any in the future, I will remove it

i18n/messages/uk.json
Iarynovskyi: it is a habit after snake_style

Mike-Popovych (approved):

IrynaKhylchuk (approved):

danikua (approved):

i18n/routing.ts
bandvov: BTW it can be done using i18n config:
```   detection: {
      // Order and from where user language should be detected
      order: ['querystring', 'cookie', 'localStorage', 'navigator', 'htmlTag'],

      // Keys or params to lookup language from
      lookupQuerystring: 'lang',
      lookupCookie: 'i18next',
      lookupLocalStorage: 'i18nextLng',

      // Cache user language on
      caches: ['localStorage', 'cookie'],
    },
```

ShadowOfTheSpace: > BTW it can be done using i18n config:
> 
> ```
>       // Order and from where user language should be detected
>       order: ['querystring', 'cookie', 'localStorage', 'navigator', 'htmlTag'],
> 
>       // Keys or params to lookup language from
>       lookupQuerystring: 'lang',
>       lookupCookie: 'i18next',
>       lookupLocalStorage: 'i18nextLng',
> 
>       // Cache user language on
>       caches: ['localStorage', 'cookie'],
>     },
> ```

I guess each library has its own pros and cons. However, for Next.js, I found that the combination of `next-intl` with custom middleware is a great choice. No extra client logic (translation can be used in server and client components with no "use client" transformations), no problems with static rendering, and the ability to have type safety even in translations (with some tricks, but anyway it's possible).

app/[lang]/[...unknown-route]/page.tsx
Comment (ShadowOfTheSpace) [inline app/[lang]/[...unknown-route]/page.tsx:1]: File is not formatted properly (no extra line in the end)

app/[lang]/layout.tsx
Comment (ShadowOfTheSpace) [inline app/[lang]/layout.tsx:1]: File is not formatted

app/[lang]/layout.tsx
```diff
@@ -26,23 +28,32 @@ export const metadata: Metadata = {
...
+
+export default async function RootLayout({
   children,
-}: Readonly<{
-    children: React.ReactNode;
-}>) {
+  params
+}: RootLayoutParams) {
+  const { lang } = await params;
+
```
Comment (ShadowOfTheSpace) [inline app/[lang]/layout.tsx:46]: I'd add check like [this](https://next-intl.dev/docs/getting-started/app-router/with-i18n-routing#layout) here, just to be sure. And please, check that correct not found pages is called (translated one)

app/[lang]/not-found.tsx
Comment (ShadowOfTheSpace) [inline app/[lang]/not-found.tsx:1]: File is not formatted

app/[lang]/not-found.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+  getTranslations: jest.fn().mockResolvedValue((key: string) => {
+    const translations: Record<string, string> = {
+      '404_language': '404 Language',
+      'goHome': 'go home',
+    };
+    return translations[key];
+  })
+}));
+
+describe('NotFound ', () => {
```
Comment (ShadowOfTheSpace) [inline app/[lang]/not-found.test.tsx]:
```suggestion
describe('NotFound', () => {
```

i18n/messages/en.json
```diff
@@ -0,0 +1,20 @@
...
+    "copyright": "© 2025 Liotoshynsky Foundation. All rights reserved.",
+    "linkPrivacy": "Privacy Policy",
+    "linkTerms": "Terms of Use",
+    "linkMedia": "Information for Media / Partners"
+  },
+  "home": {
+    "text": "Liatoshynsky project"
+  },
+  "common": {
+    "404_language": "The page is not yet available in this language",
```
ShadowOfTheSpace: Can you use the camel case here? Just to be consistent with other translations. For example `languageNotFound`

i18n/messages/en.json
```diff
@@ -0,0 +1,20 @@
...
+    "404_page": "404 Page not found",
```
ShadowOfTheSpace: I'd sugges to change key here, it's not clear what is that. Is it a tab title or is it a text on the screen?
Try to sync with other devs and think about translation keys system. I'l send you an example in DM in Discord.

i18n/request.ts
```diff
@@ -0,0 +1,17 @@
...
+export default getRequestConfig(async ({requestLocale}) => {
+  const requested = await requestLocale;
+  const locale = hasLocale(routing.locales, requested)
+    ? requested
+    : routing.defaultLocale;
+
+  return {
+    locale,
+    messages: (await import(`./messages/${locale}.json`)).default,
+    onError: () => {notFound();}
```
ShadowOfTheSpace: Does it work?
```suggestion
    onError: notFound
```

middleware.ts
```diff
@@ -1,12 +1,8 @@
...
-export const middleware = (request: NextRequest) => {
-  return urlLocaleMiddleware(request);
-};
+export default createMiddleware(routing);
 
 export const config = {
-  matcher: [
-    '/((?!api|_next/static|_next/image|favicon.ico|.*\\.[^/]+$).*)'
-  ],
+  matcher: '/((?!api|_next/static|_next/image|favicon.ico|.*\\.[^/]+$).*)'
```
ShadowOfTheSpace: Probably, we can exclude `favicon.ico` since we have matcher to exclude all files. But I'm not sure about that, so check it pls.
```suggestion
  matcher: '/((?!api|_next/static|_next/image|.*\\.[^/]+$).*)'
```

app/[lang]/layout.tsx
Comment (Iarynovskyi) [reply app/[lang]/layout.tsx:46]: All works correct 
Video do not downloading so screen :P
But it works 

Знімок екрана 2025-05-30 о 18 10 04

i18n/request.ts
Iarynovskyi: `    onError: () => notFound()
`

if i write like this it is working

https://github.com/user-attachments/assets/43e6f362-c023-40cd-9f09-9b677aa94c37

middleware.ts
Iarynovskyi: seems to work good, so i will delete this

PR #86, opened 2025-05-24
Middleware meddling with icon imports fix 

### LINKED ISSUE #84: Middleware changing request path meddling with icon imports
## Reproducible when

always

## Reproducible by

Opening site's page

## Description

When opening site icons do not load

Appearence:

Image

Browser console errors:

Image

 ## Possible cause

As described in error, site cannot find [lang]/[icon name].svg icon, this is most possible caused by locale adding middleware changing requests for static resources

### LINKED ISSUE #114: Small middleware matcher Regex error
## Current Regex:

\/((?!api|_next\/static|_next\/image|favicon.ico|.\*\\.[^\/]+$).*)

## Visualization:

Image

## Description

Because '.' character between favicon and ico isn't treated as character by Regex, but as control character which means "any character", its possible to create strange paths that wont go through middleware like /favicon#ico. This is not severe as only one character can be inbetween favicon and ico, but issue nonetheless

yur4uwe:
## Summary of issue

Solution for bug #84. 

## Summary of change

Solves issue by changing middleware matcher Regex for Next config. New regex additionaly matches all paths which have '.' and at least 1 non-slash character after, up to the end of the string

### Before fix

image

### After fix

image

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -7,6 +7,6 @@ export const middleware = (request: NextRequest) => {
 
 export const config = {
     matcher: [
-        '/((?!api|_next/static|_next/image|favicon.ico).*)'
+        '/((?!api|_next/static|_next/image|favicon.ico|.*\\.[^/]+$).*)'
     ],
 };
--- a/app/lib/i18n/utils.test.ts
+++ b/app/lib/i18n/utils.test.ts
@@ -1,6 +1,7 @@
 import {
     extractLocaleFromPath,
-    determineLocale
+    determineLocale,
+    isIcon
 } from './utils';
 import {
     SUPPORTED_LOCALES,
@@ -36,4 +37,19 @@ describe('utils tests', () => {
             expect(determineLocale('')).toBe(DEFAULT_LOCALE);
         });
     });
+
+    describe('isIcon', () => {
+        it('should return true for paths with file extensions', () => {
+            expect(isIcon('/favicon.ico')).toBe(true);
+            expect(isIcon('/images/logo.png')).toBe(true);
+            expect(isIcon('/styles/main.css')).toBe(true);
+        });
+
+        it('should return false for paths without file extensions', () => {
+            expect(isIcon('/about')).toBe(false);
+            expect(isIcon('/')).toBe(false);
+            expect(isIcon('.')).toBe(false);
+        });
+    })
+
 });
--- a/app/lib/i18n/utils.test.ts
+++ b/app/lib/i18n/utils.test.ts
@@ -1,7 +1,6 @@
 import {
     extractLocaleFromPath,
-    determineLocale,
-    isIcon
+    determineLocale
 } from './utils';
 import {
     SUPPORTED_LOCALES,
@@ -37,19 +36,4 @@ describe('utils tests', () => {
             expect(determineLocale('')).toBe(DEFAULT_LOCALE);
         });
     });
-
-    describe('isIcon', () => {
-        it('should return true for paths with file extensions', () => {
-            expect(isIcon('/favicon.ico')).toBe(true);
-            expect(isIcon('/images/logo.png')).toBe(true);
-            expect(isIcon('/styles/main.css')).toBe(true);
-        });
-
-        it('should return false for paths without file extensions', () => {
-            expect(isIcon('/about')).toBe(false);
-            expect(isIcon('/')).toBe(false);
-            expect(isIcon('.')).toBe(false);
-        });
-    })
-
 });
--- a/app/lib/i18n/utils.ts
+++ b/app/lib/i18n/utils.ts
@@ -19,3 +19,7 @@ export const determineLocale = (localeFromCookie: string): string => {
     }
     return DEFAULT_LOCALE;
 };
+
+export const isIcon = (pathname: string): boolean => {
+    return /\.[^/]+$/.test(pathname);
+}
--- a/app/lib/i18n/utils.ts
+++ b/app/lib/i18n/utils.ts
@@ -21,5 +21,9 @@ export const determineLocale = (localeFromCookie: string): string => {
 };
 
 export const isIcon = (pathname: string): boolean => {
-    return /\.[^/]+$/.test(pathname);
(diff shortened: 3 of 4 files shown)


Iarynovskyi:
https://github.com/user-attachments/assets/54029add-7d97-4ca2-ba64-3b88ada96d24

yur4uwe:
> 2025-05-24.18.44.19.mov

I see the issue

Iarynovskyi:
> > 2025-05-24.18.44.19.mov
> 
> Is this footage before or after applying my changes?

After. I yesterday looked on this bug so know where to check it

app/components/Footer/FooterCopyrights/FooterCopyrights.styles.ts

In this file change icon import to this one    ->    backgroundImage: 'url(/Bullet-icon.svg)',
Without ' . ' and than it should work correct

yur4uwe:
> > > 2025-05-24.18.44.19.mov
> > 
> > 
> > Is this footage before or after applying my changes?
> 
> After. I yesterday looked on this bug so know where to check it
> 
> app/components/Footer/FooterCopyrights/FooterCopyrights.styles.ts
> 
> In this file change icon import to this one -> backgroundImage: 'url(/Bullet-icon.svg)', Without ' . ' and than it should work correct

Its sure isnt the case because i use next Image component for another icon component with /copy-icon.svg, and it still fails to fetch it

yur4uwe:
image

here is without isIcon function

Iarynovskyi:
> image
> 
> here is without isIcon function

Well, I just checked the icons that are shown in bug here and with your function and removing ' . ' it was fixed, but if your icon in your PR does not work, then the problem is deeper..

yur4uwe:
> > image
> > here is without isIcon function
> 
> Well, I just checked the icons that are shown in bug here and with your function and removing ' . ' it was fixed, but if your icon in your PR does not work, then the problem is deeper..

Well, that why i created this pr after all

bandvov:
> > > image
> > > here is without isIcon function
> > 
> > 
> > Well, I just checked the icons that are shown in bug here and with your function and removing ' . ' it was fixed, but if your icon in your PR does not work, then the problem is deeper..
> 
> Well, that why i created this pr after all

Lets first merge PR with i18n setup(#83 ). After that you should check if problem is solved

ShadowOfTheSpace (commented):
Can you enhance the `matcher` of the middleware instead of all this extra logic?

Iarynovskyi:
> Can you enhance the `matcher` of the middleware instead of all this extra logic?

The middleware `matcher` does not apply to files in the` public/` folder
They are served statically and bypass middleware
So i think we need extra logic for handling those cases

yur4uwe:
> Can you enhance the `matcher` of the middleware instead of all this extra logic?

For sure, the only reason i added this custom logic comes from personal preferences and lacking understanding of regex

yur4uwe:
> The middleware `matcher` does not apply to files in the` public/` folder They are served statically and bypass middleware So i think we need extra logic for handling those cases

` public/` folder do get through middleware though, thats why this bug exists in the first place

yur4uwe:
> Can you enhance the `matcher` of the middleware instead of all this extra logic?

I sure can, its just a matter of my taste to add string manipulation, as i understand it better, but i see the reason for them to be added

ShadowOfTheSpace:
> > Can you enhance the `matcher` of the middleware instead of all this extra logic?
> 
> The middleware `matcher` does not apply to files in the` public/` folder They are served statically and bypass middleware So i think we need extra logic for handling those cases

Probably you have problems with `matcher`. Please, check on which routes `middleware` invokes. Just add console.log to the `req.nextUrl` or smth like that inside body of the middleware (middleware.ts file). If there are routes that should be ignored by the matcher - something is wrong.
Also, check [this article on the Nextjs docs](https://nextjs.org/docs/app/building-your-application/routing/middleware#matching-paths), middleware invokes on every route (it doesn't matter whether the route is static or not).

Also in the `next-intl` lib, I saw that they are [proposing](https://next-intl.dev/docs/getting-started/app-router/with-i18n-routing#middleware) to include `.*\\..*` into the matcher (it will exclude all URLs that look like files, for example - static files). It might help with this issue.

ShadowOfTheSpace:
> > Can you enhance the `matcher` of the middleware instead of all this extra logic?
> 
> I sure can, its just a matter of my taste to add string manipulation, as i understand it better, but i see the reason for them to be added

Of course, it's better to do what you understand, but in this case - better to try understand middleware `matchers` instead of additional logic inside middleware. 

Remember, that NextJS middleware - is a thing that runs on every request. **More logic on every request - so they will be slower.** Do we want that? Probably not)

Instead of additional logic, you can use `matcher` that was designed for those cases by NextJS devs to skip middleware invocation. It's the fastest (by performance) possible approach, so don't be afraid to use something new, it's worth it! 🚀

Iarynovskyi (approved):

middleware.ts
```diff
@@ -7,6 +7,6 @@ export const middleware = (request: NextRequest) => {
 
 export const config = {
     matcher: [
-        '/((?!api|_next/static|_next/image|favicon.ico).*)'
+        '/((?!api|_next/static|_next/image|favicon.ico|.*\\.[^/]+$).*)'
```
ShadowOfTheSpace, line 10: Nice modification with `.*\\.[^/]+$`.
Is it really work for `route.with.dot/example`?

ShadowOfTheSpace (approved):

middleware.ts
yur4uwe: just checked, and doesn't work as expected, investigating this issue

yur4uwe: False alarm, server wasn't working

regex works as expected (lets middleware process it) with this path

ShadowOfTheSpace: So it skips static files, but runs for the routes (even with a dot in the path), right?

yur4uwe: exactly

though it wont work for something like /example.smth.anyth and /example.smth.

oleg191006 (approved):

danikua (approved):

PR #87, opened 2025-05-24
(SP: 2) Implement Backend Logic for Contact Form Submission

### LINKED ISSUE #35: (SP: 2) Implement Backend Logic for Contact Form Submission
Develop the backend API route in Next.js to handle contact form submissions.

This complements the frontend validation logic and submission trigger. On successful submission, the backend should process the form data appropriately
You should:
1.Create a POST API route

2.Accept JSON payload with:

- name: string (2–50 characters)
- email: string (valid email)
- message: string (10–1000 characters)

3.Validation (Server-side):
 Return 400 status with error message for invalid data
**On Successful Submission:**
Return 200 OK with { success: true }
Optioonal(for now)

1. Store the message in a database
2. Send an email notification

oleg191006:
## Summary of change
Added a backend route /api/contact to handle contact form submissions
It accepts POST requests with a JSON payload (name, email, message), performs server-side validation, and returns either validation errors (400) or { success: true } (200)

## Result

https://github.com/user-attachments/assets/10839bc4-0036-4f1b-a434-46eb2292fa53

### DIFF
--- a/app/api/contact/route.ts
+++ b/app/api/contact/route.ts
@@ -1,0 +1,13 @@
+import { NextResponse } from 'next/server';
+import { validateContactData } from '~/lib/utils/validateContactData';
+
+export async function POST(request: Request) {
+  const data = await request.json();
+  const errors = validateContactData(data);
+
+  if (errors.length > 0) {
+    return NextResponse.json({ success: false, errors }, { status: 400 });
+  }
+
+  return NextResponse.json({ success: true });
+}
--- a/app/api/contact/route.ts
+++ b/app/api/contact/route.ts
@@ -1,4 +1,4 @@
-import { NextResponse } from 'next/server';
+import { successResponse, errorResponse } from '~/lib/utils/apiResponse';
 import { validateContactData } from '~/lib/utils/validateContactData';
 
 export async function POST(request: Request) {
@@ -6,8 +6,8 @@ export async function POST(request: Request) {
   const errors = validateContactData(data);
 
   if (errors.length > 0) {
-    return NextResponse.json({ success: false, errors }, { status: 400 });
+    return errorResponse(errors, 400);
   }
 
-  return NextResponse.json({ success: true });
+  return successResponse({ data, success: true });
 }
--- a/app/api/contact/route.ts
+++ b/app/api/contact/route.ts
@@ -1,7 +1,9 @@
 import { successResponse, errorResponse } from '~/lib/utils/apiResponse';
-import { validateContactData } from '~/lib/utils/validateContactData';
+import {
+  validateContactData,
+  type ContactFormData,
+} from '~/lib/utils/validateContactData';
 import { validateRequestData } from '~/lib/utils/validateRequestData';
-import { ContactFormData } from '~/lib/utils/validateContactData';
 
 export async function POST(request: Request) {
   const data = await request.json();
--- a/app/api/contact/route.ts
+++ b/app/api/contact/route.ts
@@ -1,13 +1,21 @@
 import { successResponse, errorResponse } from '~/lib/utils/apiResponse';
 import { validateContactData } from '~/lib/utils/validateContactData';
+import { validateRequestData } from '~/lib/utils/validateRequestData';
+import { ContactFormData } from '~/lib/utils/validateContactData';
 
 export async function POST(request: Request) {
   const data = await request.json();
-  const errors = validateContactData(data);
 
-  if (errors.length > 0) {
-    return errorResponse(errors, 400);
+  const validationResult = validateRequestData<ContactFormData>(
+    data,
+    validateContactData,
+  );
+
+  if (!validationResult.valid) {
+    return errorResponse(validationResult.errors);
   }
 
-  return successResponse({ data, success: true });
+  const { value } = validationResult;
+
+  return successResponse({ data: value, success: true });
 }
--- a/app/constants/validation.ts
+++ b/app/constants/validation.ts
@@ -1,0 +1,17 @@
+export const lengths = {
+  NAME_MIN_LENGTH: 2,
+  NAME_MAX_LENGTH: 50,
+  MESSAGE_MIN_LENGTH: 10,
+  MESSAGE_MAX_LENGTH: 1000,
+};
+
+export const regex = {
+  EMAIL_PATTERN:
+    /^([a-z\d]+([._-][a-z\d]+)*)@([a-z\d]+([.-][a-z\d]+)*\.[a-z]{2,})$/i,
+};
+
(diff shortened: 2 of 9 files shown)


app/constants/validation.ts
```diff
@@ -0,0 +1,17 @@
...
+export const regex = {
+  EMAIL_PATTERN:
+    /^([a-z\d]+([._-][a-z\d]+)*)@([a-z\d]+([.-][a-z\d]+)*\.[a-z]{2,})$/i,
+};
+
+export const error = {
+  NAME_ERROR: `Name must be a string between ${lengths.NAME_MIN_LENGTH} and ${lengths.NAME_MAX_LENGTH} characters long.`,
+  EMAIL_ERROR: 'Invalid email address.',
+  MESSAGE_ERROR: `Message must be a string between ${lengths.MESSAGE_MIN_LENGTH} and ${lengths.MESSAGE_MAX_LENGTH} characters.`,
+};
```
luvthenika: i think it would be a good practice to separate error messages object into a dedicated error file, where all error message may be stored

app/api/contact/route.ts
```diff
@@ -0,0 +1,13 @@
+import { NextResponse } from 'next/server';
+import { validateContactData } from '~/lib/utils/validateContactData';
+
+export async function POST(request: Request) {
+  const data = await request.json();
+  const errors = validateContactData(data);
+
+  if (errors.length > 0) {
+    return NextResponse.json({ success: false, errors }, { status: 400 });
```
Iarynovskyi: Maybe we can do some Factory for base responses ?
Where for now we add for example `ok` and `error` and expand in future for others like `create` etc.
Because now it looks very hardcoded.

app/api/contact/route.ts
```diff
@@ -0,0 +1,13 @@
+import { NextResponse } from 'next/server';
+import { validateContactData } from '~/lib/utils/validateContactData';
+
+export async function POST(request: Request) {
+  const data = await request.json();
+  const errors = validateContactData(data);
```
Iarynovskyi: I think that all validation logic must to be not in API layer but in BL

Iarynovskyi:
**Just an idea to discuss:**

Right now, our API logic is written directly in the `route.ts` files. While it works fine for now, it might become harder to maintain, test, or scale as the project grows.

What do you think about separating responsibility into controllers, services, and repositories (DAL), with DI between them ?

This structure would help us:
- write more modular and reusable code
- keep business logic decoupled from database logic
- simplify testing by injecting mock dependencies
- make the backend easier to maintain and evolve as requirements grow

It is not about adding complexity, but just setting up a clean foundation that can grow with the project and support better backend practices.

app/api/contact/route.ts
bandvov: In this case it is validation for request fields so it should be here.

oleg191006:
> **Just an idea to discuss:**
> 
> Right now, our API logic is written directly in the `route.ts` files. While it works fine for now, it might become harder to maintain, test, or scale as the project grows.
> 
> What do you think about separating responsibility into controllers, services, and repositories (DAL), with DI between them ?
> 
> This structure would help us:
> 
> * write more modular and reusable code
> * keep business logic decoupled from database logic
> * simplify testing by injecting mock dependencies
> * make the backend easier to maintain and evolve as requirements grow
> 
> It is not about adding complexity, but just setting up a clean foundation that can grow with the project and support better backend practices.

I totally agree with you that this structure will definitely help with scalability, maintainability and testing in the long run. However, at this stage the business logic is still not defined, so it is not entirely clear what exactly will go into the service or controller layer. Once things become clearer, then we can properly distribute the logic

Iarynovskyi (approved):

danikua (approved):

app/lib/utils/apiResponse.ts
```diff
@@ -0,0 +1,17 @@
+import { NextResponse } from 'next/server';
+
+interface ErrorResponse {
+  success: false;
+  errors: string[] | Record<string, unknown>;
+}
+
+export const successResponse = <T>(data: T, status: number = 200) => {
```
ShadowOfTheSpace: If you don't use the `data` parameter inside this generic function and just pass it into the `.json` function, you can omit the generic at all. 
```suggestion
export const successResponse = (data: unknown, status: number = 200) => {
```

app/api/contact/route.ts
```diff
@@ -0,0 +1,13 @@
+import { successResponse, errorResponse } from '~/lib/utils/apiResponse';
+import { validateContactData } from '~/lib/utils/validateContactData';
+
+export async function POST(request: Request) {
+  const data = await request.json();
+  const errors = validateContactData(data);
+
+  if (errors.length > 0) {
+    return errorResponse(errors, 400);
```
ShadowOfTheSpace: `400` is by default, isn't it?
```suggestion
    return errorResponse(errors);
```

app/lib/utils/validateContactData.ts
```diff
@@ -0,0 +1,35 @@
+import { lengths, regex } from '~/constants/validation';
+import { errors } from '~/constants/errors';
+
+export type ContactFormData = {
+  name: unknown;
+  email: unknown;
+  message: unknown;
+};
+
+export function validateContactData(data: ContactFormData): string[] {
```
ShadowOfTheSpace, line 10: I'd also suggest creating a generic `validateRequestData` or `parseRequestData` function that will look something like this:

```ts
const validateRequestData = <T>(
	data: T,
	validationFunction: (data: T) => string[],
): { errors: string[]; valid: false } | { valid: true; value: T } => {
	const errors = validationFunction(data);

	return errors.length === 0
		? { valid: true, value: data }
		: { errors, valid: false };
};
```

And usage of that will be looked like this (in API routes)

```ts

const validationResult = validateRequestData<ContactFormData>(data, validateContactData);

if(!validationResult.valid){
  return errorResponse(validationResult.errors)
}

const { value } = validationResult

```

It's just a pseudocode (probably there are some type or generic errors), but I think you got the idea

ShadowOfTheSpace (approved):
Thanks for applying my suggestions!
Also, just a question, do you consider using only manual validation or you will add a validation lib, like Zod?

app/lib/utils/validateContactData.ts
oleg191006: that's a good idea, thanks, I've already done it

oleg191006:
> Thanks for applying my suggestions! Also, just a question, do you consider using only manual validation or you will add a validation lib, like Zod?

We haven't discussed this yet, but I'll raise this issue at the meeting tomorrow

PR #90, opened 2025-05-26
Implemented Footer contact info block

### LINKED ISSUE #22: (SP: 1) Add redirection to users email client on foundations email click
Integrate the contact email address into the footer with a clickable link that opens the user's default email client with the address pre-filled. 

Image

- [x] On hover, cursor should be changed into a pointer
- [x] On click, user should be redirected to email client
- [x] After redirect, user should have letter sending window open with "receiver" field filled with foundations email

yur4uwe:
## Summary of issue

Created component for display of company contacts

- [x] Email link redirects to `mailto:<email>` opening default mail client
- [x] Styles are structured according to mockup
- [x] Telephone link copies number when using desktop
- [x] Telephone link uses `tel` protocol on mobile devices  

### Desktop

https://github.com/user-attachments/assets/e6c3baa1-6359-4a63-8a6a-db757cf369d4

### Phone

https://github.com/user-attachments/assets/eb762b92-a7f4-4f98-8f04-f4896316c7c7

## Design

Mockup:

image

Implementation:

image


IrynaKhylchuk:
I noticed that some lines use 4-space indentation instead of 2, which is what we have in our [prettier.config.js](https://github.com/Liatoshynsky-Foundation/lf-client/blob/7e473dfe7828556fe77f440f9bd78f7fb39efb68/prettier.config.js#L4).
Could you please update the indentation?
Also, could you take a look at the SonarQube issue and resolve it?

app/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -3,28 +3,28 @@ import CustomCheckbox from './Checkbox';
...
-        render(<CustomCheckbox disabled label={label} />);
-        expect(screen.queryByRole('checkbox')).not.toBeChecked();
-    });
-    it('should display checked checkbox by default', async () => {
-        render(<CustomCheckbox label={label} defaultChecked />);
-        expect(screen.queryByRole('checkbox')).toBeChecked();
-    });
+  });
+  it('should not toggle disabled checkbox', async () => {
+    render(<CustomCheckbox disabled label={label} />);
```
IrynaKhylchuk: What do you think about moving the `<CustomCheckbox label={label} defaultChecked />` component rendering into a beforeEach?

IrynaKhylchuk (approved):
Approved. Please take a look at my comment on [Checkbox.spec.tsx](https://github.com/Liatoshynsky-Foundation/lf-client/pull/90/files#r2113624331) before merging.

app/components/Checkbox/Checkbox.spec.tsx
yur4uwe: That would be an awsome idea. If i were to be an author of this test I would definitely use it but not for this particular test though. You see, it requires additional attribute which would be unnecessary hard to pull off when render is in beforeEach. But for more insight i recommend you ask @luvthenika, author of said test, what would be better

IrynaKhylchuk: Sorry, I didn't notice that you are not the author of this test.

IrynaKhylchuk (approved):

bandvov (approved):

Iarynovskyi (approved):

PR #111, opened 2025-05-26
Added all svg icons

### LINKED ISSUE #110: (SP: 1) Import all svg icons
Image

luvthenika:


bandvov (approved):

VKormylo:
There are a few files that contain spaces in their name, I think it's better to rename them including hyphen for example.

yur4uwe:
In my humble opinion they are little too grey. I think we should ask designers about chaging color of those that should be darker

IrynaKhylchuk:
Just to clarify — aren’t we using the 'Lucide' library for icons?

uliaescha (approved):

public/Frame.svg
lizabre, line 1: Weren't we supposed to upload a 'lucid' package for all icons that were not made by our design team and are available there?

public/Bullet small.svg
lizabre, line 1: Maybe it would be better to put it in a nested folder like 'icons', as Maria menioned

ShadowOfTheSpace: Yep, you should add a common folder for all icons.

Additionally, please check other variants of icon management in Next.js. I feel that a public folder is not the best choice for SVG icons that can be mounted directly in HTML, rather than being served statically from the Next.js server.

I almost always used libs like [Lucide](https://lucide.dev/) or other icon libraries. 
But even when I have my icons, it always ends up with structure like [this](https://github.com/ShadowOfTheSpace/jazz/blob/main/src/libs/components/icon/libs/maps/icon-name-to-svg/icon-name-to-svg.map.ts) and dedicated [Icon component](https://github.com/ShadowOfTheSpace/jazz/blob/main/src/libs/components/icon/icon.tsx) that used this icon map (icons can be imported from files or component libs).

The example above is from a React project, but I think this approach can be used in Next.js as well.

Each approach has its own pros and cons, so better to discuss that with the team ↓

- embedded icons into HTML increase the size of the page, but icons are already ready for the user after the page is loaded (no extra request for each icon)
- icons in the public folder decrease the size of the page, but the user needs to wait for each icon to be loaded from the server (I think it's a bad thing from a UX perspective). Also, in an "ideal" world - static files are cached on CDN and not served from the server itself and I think that you not gonna do that. So this approach just increases server load, because instead of just managing pages it should be like CDN.

ShadowOfTheSpace:
> Just to clarify — aren’t we using the 'Lucide' library for icons?

If that is true about the usage of the `Lucide` lib, my comment ↑ makes even more sense, so discuss that with the team, please.

public/Bullet small.svg
bandvov: Probably there will be CDN on hosting. Not sure 100% though

Iarynovskyi (approved):

PR #130, opened 2025-05-28
Implemented 'PersonalCard' element

### LINKED ISSUE #66: (SP: 2) Create a person card element aligned with design system
Implement person card for mobile and desktop screens
1. The view on desktop 

Image

2. The view on mobile 

Image

It should consists of:

- [x] - photo bubble
- [x] - title 
- [x] - description
- [x] - logo

IrynaKhylchuk:
**Implemented 'PersonalCard' element.**

The card includes:
- stylized photo bubble;
- title (name);
- description;
- logo.  

Desktop version:
Screenshot 2025-05-28 131606

Mobile version:
Screenshot 2025-05-28 131617

Tests:
Screenshot 2025-05-28 130405


luvthenika (approved):

danikua (approved):

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { Box, Typography } from '@mui/material';
+import { styles } from './PersonCard.styles';
+import Logo from '../../../../public/Logo.svg';
+import Image from 'next/image';
+
+interface PersonCardProps {
+  photo: string;
```
ShadowOfTheSpace: I'd suggest using a more meaningful name here, like `imageUrl`, `photoUrl`, or `picUrl` because it's not totally clear that it should be a link.
Also, do you have guys some conventions for image URLs prop names? It will be hard to manage if all props + fields in DB will have inconsistent names.

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+  name: string;
+  description: string;
+}
+
+const PersonCard = ({ photo, name, description }: PersonCardProps) => {
```
ShadowOfTheSpace: What about the `React.FC` type that is used in other components? Do you have a consistent template for React components?

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <Box sx={styles.container}>
+      <Box sx={styles.cardContent}>
+        <Box sx={styles.photoWrapper}>
+          <Image
+            alt="Person card photo"
```
ShadowOfTheSpace: It's not recommended to add `photo`, `pic`, or `image` words to the `alt` attribute. Check that out: https://help.siteimprove.com/support/solutions/articles/80000863904

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+            width={185}
+            height={166}
+            style={{
+              objectFit: 'cover',
+              borderRadius: '60% 40% 60% 40% / 55% 45% 55% 45%',
+              position: 'absolute',
+              top: '2px',
+              left: '15px',
+              transform: 'rotate(25deg)',
+            }}
```
ShadowOfTheSpace: Can we extract that into `styles` file?

public/Logo.svg
ShadowOfTheSpace, line 1: Don't we already have the logo image in the `public` folder or it's a different one?

IrynaKhylchuk: When I created the PR, there was no logo in the public folder yet.
The one that’s there now has a different color and size, so I used this version that matches the design.

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Thanks for the advice! I moved the styles to a `styles` file.

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Thanks for the suggestion! I’ve updated the name to make it more meaningful.
As for the database, as far as I know, we don't have any established conventions for image URL prop names yet.

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Thanks for the suggestion! I’ve updated the `alt` attribute.

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Thanks for the suggestion! I’ve added `React.FC` to this component.

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { Box, Typography } from '@mui/material';
+import { styles } from './PersonCard.styles';
+import Logo from '../../../../public/Logo.svg';
```
ShadowOfTheSpace: Can we use import alias here? Like `~/public/Logo.svg`? As I see [here](https://github.com/Liatoshynsky-Foundation/lf-client/blob/e0739175cf49ff6630d1208d72bac3b2b7e148e7/tsconfig.json#L23), it's possible

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+            style={styles.image}
+          />
+        </Box>
+        <Box sx={styles.textWrapper}>
+          <Typography sx={styles.name}>{name}</Typography>
+          <Typography sx={styles.description}>{description}</Typography>
+        </Box>
+      </Box>
+      <Box sx={styles.logoWrapper}>
+        <Box component="img" src={Logo.src} alt="Logo" sx={styles.logo} />
```
ShadowOfTheSpace: What about the `alt` attribute here? Will be possible to translate it with this approach?

app/components/design-system/PersonCard/PersonCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+const PersonCard: React.FC<PersonCardProps> = ({
+  imgURL,
+  name,
+  description,
+}) => {
+  return (
+    <Box sx={styles.container}>
+      <Box sx={styles.cardContent}>
+        <Box sx={styles.photoWrapper}>
+          a
```
ShadowOfTheSpace: I guess it's redundant
```suggestion
```

public/Logo.svg
ShadowOfTheSpace: If the logo from the current `public` folder is different from your needs, you should name a new one with a different name (in the same `public/images` folder), for example, `gray-logo` or smth like that. 
Try to use as many meaningful names, as possible.

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Yes, we already have the i18n library set up, but we haven't added any translations yet. I plan to include them in the future.

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Yes, I’ve changed it to use the alias. Thank you for pointing that out!

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: Removed — thanks for noticing!

app/components/design-system/PersonCard/PersonCard.tsx
IrynaKhylchuk: I changed the import to use the path `/images/light-logo.svg` directly in the `src` attribute because Next.js does not allow importing files (including SVGs) directly from the `public` folder using import. Files in the `public` folder should be referenced by their URL path instead.

ShadowOfTheSpace (approved):

PR #136, opened 2025-05-29
Implemented button component

### LINKED ISSUE #45: (SP: 3) Create Button element aligned with design system
Create a Button element for universal usage through the design system based on designs from Figma

Requirements:

- should follow the design based on the current mock-ups in Figma 
- should be usable universally across the app

Props:
- size
- color
- variant (filled, outlined, text)
- states (enabled, hovered, focused, pressed, disabled)

Image

uliaescha:
## Summary of issue

Create Button element aligned with design system 
image

## Summary of change
Added custom props such as: color, size, variant due to design system 
Primary button: small, medium, large, filled, outlined, text 

image
image

Secondary button: small, medium, large, filled, outlined, text 

image
image

Tertiary button: small, medium, large, filled 
 
image
image

Test coverage: 
image


irynalaitaruk:
@uliaescha, please fix Sonar issues

Iarynovskyi:
> @uliaescha, please fix Sonar issues

And do not forgot to fix eslint errors

app/constants/types/common.enum.ts
ShadowOfTheSpace, line 1: Why do you need enums here if you defined types of props in the component?

app/components/Button/Button.tsx
```diff
@@ -0,0 +1,100 @@
...
+} from '@mui/material';
+
+import {
+  buttonBaseStyles,
+  variantStyles,
+  sizeStyles,
+  typographyStyles
+} from './Button.styles';
+
+const sizes = ['large', 'medium', 'small'] as const;
```
ShadowOfTheSpace: If you want to use your enum as a value of prop and still have the ability to pass a plain string, you can do this:

```ts
type Size = `${ButtonSize}`
```

Or if you don't need enums for that - just have `type Size = 'large' | 'medium' | 'small'` - simple union type

The less code you have after TS to JS conversion - the better. Because in JS your types are gone, but the `size` array is still present and unused (check this [ts-playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBBCWAvAphGBeGBtA5AGwEMAnAc2RwBoYcBbZAE3gFcbLqIaC88cBdGAmlCQoAbgBQ44dBgAHIiFkBRAB4EasvMgBcMABRQAnrOQgAZnCSoAlFjAsARsiL9MAIkKlkb0UA)).

uliaescha: Thanks! That makes total sense. I removed the unused sizes and variants arrays and switched to union types

lizabre (approved):

danikua (approved):

bandvov (approved):

ShadowOfTheSpace (approved):

Iarynovskyi (approved):

luvthenika (approved):

PR #147, opened 2025-05-30
Feature/55/tooltip component

### LINKED ISSUE #55: (SP: 3) Create Tooltip element aligned with design system
Create a Tooltip element for universal usage through the design system based on designs from Figma

Requirements:

should follow the design based on the current mock-ups in Figma
should be usable universally across the app

Image

irynalaitaruk:
Implemented tooltip component #55
Added unit tests
Screenshot 2025-05-30 at 16 42 14
Screenshot 2025-05-30 at 16 42 39

Screenshot 2025-05-30 at 16 18 40

### DIFF
--- a/app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
+++ b/app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
@@ -3,7 +3,7 @@ export const tooltipStyles = {
   color: '#ffffff',
   fontSize: '14px', 
   fontStyle: 'italic',
-  fontFamily: 'var(--font-oswald)',
+  fontFamily: 'var(--font-mulish)',
   borderRadius: '18px',
   padding: '8px 20px',
 };
--- a/app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
+++ b/app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
@@ -3,6 +3,7 @@ export const tooltipStyles = {
   color: '#ffffff',
   fontSize: '14px', 
   fontStyle: 'italic',
+  fontFamily: 'var(--font-oswald)',
   borderRadius: '18px',
   padding: '8px 20px',
 };
--- a/app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
+++ b/app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
@@ -1,5 +1,5 @@
 export const tooltipStyles = {
-  backgroundColor: '#52545F',
+  backgroundColor: '#52545A',
   color: '#ffffff',
   fontSize: '14px', 
   fontStyle: 'italic',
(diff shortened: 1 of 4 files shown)


app/shared/components/design-system/all-components/tooltip/Tooltip.styles.ts
```diff
@@ -0,0 +1,13 @@
+export const tooltipStyles = {
+  backgroundColor: '#52545F',
```
VKormylo: In our design system tooltip has this color: #52545A. So please check that, I think you mistyped that.

app/shared/components/design-system/all-components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,29 @@
...
+        title='My Tooltip'
+        placement="top"
+        arrow={showArrow}
+        componentsProps={{
+          tooltip: { sx: tooltipStyles },
+          arrow: { sx: arrowStyles },
+        }}
+      >
+        <Typography>
+          My Tooltip
```
VKormylo: You should get text of the tooltip via props rather than hardcoding it, because this component should be reusable.

VKormylo (changes requested):

app/shared/components/design-system/all-components/tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import TooltipCustom from './Tooltip';
+
+describe('TooltipCustom Component', () => {
+  test('renders the button with tooltip', () => {
+    render(<TooltipCustom showArrow={true} />);
+    const buttonElement = screen.getByText(/My Tooltip/i);
```
danikua: I think you can define it inside beforeEach() instead of repeating in every test

app/shared/components/design-system/all-components/tooltip/Tooltip.tsx
irynalaitaruk: @VKormylo done

VKormylo (approved):

Iarynovskyi (approved):

app/shared/components/design-system/all-components/tooltip/Tooltip.test.tsx
irynalaitaruk: @danikua done

danikua: sorry, i didn't notice that `showArrow` has different values, so you can delete `beforeEach `and just define render in every test

danikua (approved):

app/shared/components/design-system/all-components/tooltip/Tooltip.test.tsx
irynalaitaruk: @danikua ready

PR #149, opened 2025-05-30
(SP:3) Create Breadcrumbs element aligned with design system

### LINKED ISSUE #68: (SP:3) Create Breadcrumbs element aligned with design system
Implement a breadcrumb component , which visually represents the naviagtion path
**States to Implement:**
- default
- hovered
- pressed
- disabled
It also may be:
**Layouts to implement**

- Single – One-level breadcrumb 
- Double – Two-level breadcrumb
- Triple – Three-level breadcrumb

Image

luvthenika:
Implemented breadcrumbs component which uses Custom Link Component
Result: 
image


app/components/Breadcrumbs/BreadCrumbs.tsx
```diff
@@ -0,0 +1,53 @@
...
+          {segment}
+        </CustomLink>
+      </Box >
+    );
+  });
+
+  return (
+    <Box >
+      <Breadcrumbs aria-label="breadcrumb" separator={<SvgImage src={ChevronRightSvg} alt='' />}>
+        <Box sx={{
```
Iarynovskyi: Please, move this `sx` to `.styles` file

app/components/Breadcrumbs/BreadCrumbs.tsx
```diff
@@ -0,0 +1,53 @@
...
+          '&hover': {
+            textDecoration: 'underline'
+          }
+        }}>
+          <CustomLink path='/' startIcon={<SvgImage src={HouseSvg} alt='' />} >
```
Iarynovskyi:
```suggestion
          <CustomLink path='/' startIcon={<SvgImage src={HouseSvg} alt='Home' />} >
```

Maybe we should add `alt` text to some important icons, since our site needs to comply with WCAG AAA?

app/components/Breadcrumbs/BreadCrumbs.tsx
```diff
@@ -0,0 +1,53 @@
+'use client';
+import * as React from 'react';
+import Breadcrumbs from '@mui/material/Breadcrumbs';
+import Typography from '@mui/material/Typography';
+import { usePathname } from 'next/navigation';
+import { Box } from '@mui/material';
+import { BreadCrumbsStyles } from './BreadCrumbs.styles';
+import CustomLink from '../Link/Link';
+import HouseSvg from '../../../public/house.svg';
```
Iarynovskyi:
```suggestion
import HouseSvg from '../../../public/icons/house.svg';
```

Iarynovskyi: And Where is this icon ?
Maybe you lost it during rebase or something like this so check to it please

app/components/Breadcrumbs/BreadCrumbs.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </Typography>
+      </Box>
+    ) : (
+      <Box>
+        <CustomLink key={href} path={href}>
+          {segment}
+        </CustomLink>
+      </Box >
+    );
+  });
```
Iarynovskyi: only this part need CSR so maybe you can move to separate component and than in this file delete "use client" and do it SSR ?

I just think that since we are using NextJS, we need to minimize client components as much as possible

app/components/Breadcrumbs/BreadCrumbs.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+        (nextNavigation.usePathname as jest.Mock).mockReturnValue('/level1/level2/level3');
+        render(<CustomBreadcrumbs />);
+
+        expect(screen.getByRole('link', { name: /home page/i })).toBeInTheDocument();
+        expect(screen.getByRole('link', { name: /level1/i })).toBeInTheDocument();
+        expect(screen.getByRole('link', { name: /level2/i })).toBeInTheDocument();
+
+        expect(screen.queryByRole('link', { name: /level3/i })).not.toBeInTheDocument();
+    });
+    test('renders Typography  if the element is the last in the breadcrumb', () => {
```
Iarynovskyi, line 41:
```suggestion
    test('renders Typography if the element is the last in the breadcrumb', () => {
```

Iarynovskyi (changes requested):
And do not forgot to move `Breadcrumbs` and `Link` to `/shared/components/design-system` folder

oleg191006:
Why is the interface extracted into a separate file for one component, but defined inside the component for another?
I think we should stick to a consistent style when defining types in components. In most cases, we use the React.FC<PropsType> approach and set default prop values directly in the function parameter destructuring. 
Also, we write components using arrow functions, not regular function declarations,so please look at how other components are written and try to follow the same conventions to maintain the same writing style

app/shared/components/Breadcrumbs/BreadCrumbs.tsx
```diff
@@ -0,0 +1,46 @@
+'use client';
+import * as React from 'react';
```
oleg191006: Is this import  necessary? Looks like it's unused

app/shared/components/CustomLink/NextLink.tsx
```diff
@@ -0,0 +1,78 @@
...
+  activeClassName?: string;
+  as?: NextLinkProps['as'];
+  href: NextLinkProps['href'];
+  linkAs?: NextLinkProps['as']; // Useful when the as prop is shallow by styled().
+  noLinkStyle?: boolean;
+} & Omit<NextLinkComposedProps, 'to' | 'linkAs' | 'href'> &
+  Omit<MuiLinkProps, 'href'>;
+
+// A styled version of the Next.js Link component:
+// https://nextjs.org/docs/pages/api-reference/components/link
```
oleg191006: please remove comments

app/shared/components/Breadcrumbs/BreadCrumbs.tsx
luvthenika: yes , it is necessary as usePathname is a client-side hook from Next.js' next/navigation module, and client-side hooks require the 'use client' 
i have tried to remove it ofc , and i got an error 
image

oleg191006: I didn't mean use client, but import * as React from 'react';

oleg191006 (approved):

Iarynovskyi (approved):

danikua (approved):

PR #198, opened 2025-06-04
Button Group implementation

### LINKED ISSUE #82: (SP: 3) Create ButtonGroup element aligned with design system
The navigation tab component should:

Allow customization of:

- [ ] Tab color (especially the active tab highlight)
- [ ] Number of tabs (dynamic rendering based on passed data)

- [ ] Support smooth transitions or animations when switching between tabs.

Design reference:
Image

yur4uwe:
## Summary of issue

Create the Button Group component to align with the design system defined in Figma. It should implement responsive styles, animated indicator movement, and proper active/inactive button states.

## Summary of change

Implemented the Button Group component using MUI’s Button Group and custom styles. Integrated state management to handle active buttons and indicator animation. Supports both default and custom color schemes.

## Testing approach

Manually test by simulating click events and verifying that the correct styles are applied for active/inactive states and the indicator moves accordingly. Automated tests ensure click handlers are triggered and styles match design expectations.

image

## CHECK LIST

- [x] Code coverage >= 95%
- [x] Unit tests for button clicks and style changes
- [x] Accessibility roles and attributes verified

## Implementation demonstration

image

image


IrynaKhylchuk:
Іs the font used here the one defined in our design system?

yur4uwe:
> Іs the font used here the one defined in our design system?

no, because it was only a demonstation of the implementation font used was defualt. I looked it up and seems that its roboto, this is a nice remark nevertheless, because i noticed that buttons do not inherit fonts. Fix is on the way

app/shared/components/Header/Header.tsx
```diff
@@ -1,15 +1,12 @@
+'use client';
```
Iarynovskyi: Why you added it here ?
I do not see any reason to do CSR for all Header

yur4uwe: Ah, i added it here because i visually tested this component in header, and some things reqired it, i will delete it before merging

bandvov (approved):

app/shared/components/design-system/all-components/button-group/ButtonGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+          }}
+          onClick={() => setActiveButton(idx)}
+          sx={{
+            ...styles.defaultButton,
+            padding: paddingSize,
+            color: idx === activeButton ? selectedButtonTextColor : buttonTextColor,
+            '&:hover': {
+              backgroundColor: alpha(selectedButtonColor, 0.1)
+            },
+            '&:active': {
```
luvthenika: prefer separating styles from markup

IrynaKhylchuk (approved):
Approved, but please fix the sonar issue.

app/shared/components/design-system/all-components/button-group/ButtonGroup.tsx
```diff
@@ -0,0 +1,80 @@
...
+      const containerRect = containerRef.current.getBoundingClientRect();
+      const buttonRect = currentButton.getBoundingClientRect();
+      const left = buttonRect.left - containerRect.left;
+      const width = buttonRect.width;
+
+      setIndicatorStyle({ left, width });
+    };
+
+    updateIndicator();
+    window.addEventListener('resize', updateIndicator);
```
Mike-Popovych: Can we use ResizeObserver instead of window.addEventListener('resize') in this component to track size changes of the container or active button and update the indicator position accordingly? If so, would it also make sense to wrap the updateIndicator function in useCallback to avoid unnecessary re-creations and keep the observer’s callback stable?

yur4uwe: nice idea! Though I don't know what exactly ResizeObserver is, because i never heard of it, so i'll look into it

yur4uwe: I implemented `ResizeObserver`, but using `useCallback` will give no real performance improvement for either observer or component itself

app/shared/components/design-system/all-components/button-group/ButtonGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+      setIndicatorStyle({ left, width });
+    };
+
+    updateIndicator();
+
+    const resizeObserver = new ResizeObserver(() => {
+      updateIndicator();
+    });
+
+    if (containerRef.current) {
```
Mike-Popovych, line 54: Observe active button too — its size may change on i18n language switch.

app/shared/components/design-system/all-components/button-group/ButtonGroup.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+    class MockResizeObserver {
+      callback: ResizeObserverCallback;
+      constructor(callback: ResizeObserverCallback) {
+        this.callback = callback;
+      }
+      observe(_target: Element) {
+        this.callback([], this);
+      }
+      unobserve(_target: Element) {
+        // no-op
```
Mike-Popovych: remove this comment, please

app/shared/components/design-system/all-components/button-group/ButtonGroup.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+      }
+      disconnect() {
+        // no-op
```
Mike-Popovych: same

app/shared/components/design-system/all-components/button-group/ButtonGroup.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+      };
+
+      return {
+        ...params,
+        toJSON: () => params
+      } as DOMRect;
+    });
+  });
+
+  afterEach(() => {
```
Mike-Popovych, line 83: Reset global ResizeObserver to avoid affecting other tests

app/shared/components/design-system/all-components/button-group/ButtonGroup.tsx
yur4uwe: the way language change is implemented, it always triggers rerender, so there is really no need to add additional event listener or whatever

app/shared/components/design-system/all-components/button-group/ButtonGroup.test.tsx
yur4uwe: sorry, i just needed anything inside that function 🤷‍♂️ to not trigger sonnar issue, but i guess i'll replace it with something else

app/shared/components/design-system/all-components/button-group/ButtonGroup.test.tsx
yur4uwe: new implementation has no means to influence tests, so i left it in `beforeAll` statement

Mike-Popovych: This global mock of ResizeObserver affects the entire test environment. To avoid side effects in other test files, it should be restored in afterAll

luvthenika (approved):

app/shared/components/design-system/all-components/button-group/ButtonGroup.test.tsx
yur4uwe: okay :(

Mike-Popovych (approved):

PR #206, opened 2025-06-05
Implemented SectionTitle component

### LINKED ISSUE #196: (SP:1) Implement SectionTitle component
Implement SectionTitle component. It consists of icon on the left and title in the middle (6th column on desktop).

- [ ] Implement SectionTitle component
- [ ] Cover component with unit tests

**How it should look like:**
Image

VKormylo:
Implemented SectionTitle component. This is reusable title component which is used in sections at the top. It can be with icon or without.

**Component takes three props:**
- icon?: boolean - include icon on the left or not
- mb:? number | string - margin bottom (if not set default is 72px)
- title: string - content of the title 

**Completed tasks:**
- [x] Implemented SectionTitle component
- [x] Covered component with unit tests
- [x] Ensured it is responsive according to figma mockups

**How it looks (laptop and above):**
image

**(mobile and tablet):**
image

**Results of tests:**
image


IrynaKhylchuk (approved):

app/shared/components/section-title/SectionTitle.styles.ts
```diff
@@ -0,0 +1,30 @@
...
+export const imageSizes = (isLaptopAndAbove: boolean) => {
+  return isLaptopAndAbove
+    ? {
+        width: 32,
+        height: 30
+      }
+    : {
+        width: 22,
+        height: 20
+      };
```
Iarynovskyi:
```suggestion
  return isLaptopAndAbove ? { width: 32, height: 30 } : { width: 22, height: 20 };
```


Make this change and you will pass ESLint checks
I do not know why ESLint is currently editing ternary operators incorrectly, so I will investigate this issue later, but for now, write it in single line

VKormylo: Ok, thanks for help. Was wondering how to fix that.

danikua (approved):

app/shared/components/section-title/SectionTitle.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+const SectionTitle: React.FC<SectionTitleProps> = ({ icon = true, mb, children }) => {
+  const { isLaptopAndAbove } = useBreakpoints();
+
+  return (
+    <Box sx={styles.container(mb)}>
+      {icon && <SvgImage src="/icons/ellipse.svg" alt="ellipse" {...imageSizes(isLaptopAndAbove)} />}
+      <Typography sx={styles.title} component="h2">
+        {children}
```
Iarynovskyi: If the content is just text, i think it is better to pass it as `title` prop instead of `children`
And if you use `children` to pass some component inside a `'use client'` component, it forces everything to render on the client, which is not good for performance and SSR

VKormylo: Ok, gonna change to title prop.

VKormylo: Done.

Iarynovskyi (approved):

PR #218, opened 2025-06-09
Footer and Header data fetch

### LINKED ISSUE #43: (SP: 2)Fetch Footer Data from Backend
📋 Description
Replace hardcoded footer content with dynamic data fetched from the backend. Implement an API endpoint that returns structured footer data stored in MongoDB.

✅ Acceptance Criteria
 An API endpoint is created that returns footer data (e.g. GET /api/footer)

 Footer data is stored in MongoDB (e.g. in a footer collection)

 The frontend fetches this data either in the layout or in a dedicated component

 Optional: Localization support is handled if needed

### LINKED ISSUE #168: (SP:1) Fetch Header Data from Backend
###  Description
Update the `Header` component to load dynamic data from the backend instead of using hardcoded values.  

---

Mike-Popovych:
###  Description

This pull request introduces a structural and functional foundation for data-driven footer and header components, following Domain-Driven Design (DDD) principles.

https://github.com/user-attachments/assets/42859974-175a-46f1-9a2b-0a34ea7556c0

###  Features

-  Implemented folder structure based on **DDD** principles
-  Set up **DI container** using the `awilix` library
-  Added **new collections** to the remote database:  
  - `foundationInfos`  
  - `navigations`
-  Created **repositories** for fetching data for the footer and header
-  Developed **services** to handle footer and header logic
-  Integrated dynamic data into the **Footer component**
-  Added unit tests for:
  - Repository logic  
  - Service logic  
  - Updated tests for the server component: Footer
-  All data is fetched **based on the current locale**

###  Checklist

- [x] Follows DDD architecture
- [x] Uses dependency injection via Awilix
- [x] Collections added to remote DB
- [x] Data-driven Footer rendering
- [x] Locale-aware data fetching
- [x] Tests included and updated


app/repositories/foundation-info/foundationInfo.repository.ts
```diff
@@ -0,0 +1,70 @@
...
+import { BrandingInfo } from '~/models/foundation-info/foundationInfoBranding';
+import { ContactInfo } from '~/models/foundation-info/foundationInfoContact';
+import { PublicInfo } from '~/models/foundation-info/foundationInfoPublic';
+import { hrefSchema } from '~/validators/constants';
+import { brandingInfoSchema, contactInfoSchema, publicInfoSchema } from '~/validators/foundationInfo.schema';
+
+export const contactRepository = {
+  async getContactInfo(): Promise<ContactInfoData> {
+    const result = await ContactInfo.findOne({ slug: 'contact-info' })
+      .select('email phone contactButtonLink socialLinks')
```
Iarynovskyi:
```suggestion
      .select('email phone contactButtonLabel socialLinks')
```
Maybe this correct ? 
Because in `faundationInfoContact` model i saw this field instead of `contactButtonLink `

Or is here correct variant so change it in `faundationInfoContact` to correct one
Because I think that the link should be returned here and this comes

Знімок екрана 2025-06-09 о 18 55 16

app/services/footerService.ts
```diff
@@ -0,0 +1,40 @@
+import { Locale } from 'next-intl';
+
+import type { FooterServiceDeps } from '~/types/types/foundationInfo.type';
+
+import dbConnect from '~/db/connect';
```
Iarynovskyi: The service layer should not know anything about DB and especially not initialize it, even if it is already in the cache, so it needs to be moved either to the repository layer or to the project creation layer

app/services/footerService.test.ts
```diff
@@ -0,0 +1,117 @@
+import { Locale } from 'next-intl';
+
+import type { FooterServiceDeps } from '~/types/types/foundationInfo.type';
+
+import { createFooterService } from '~/services/footerService';
+
+jest.mock('~/db/connect', () => ({
```
Iarynovskyi: If the DB connection initialization is done inside the service layer, then testing that service requires mocking the DB connection, because the service depends on the concrete implementation of data access

However, if you move the DB connection logic to the repository layer — which is responsible for data access — then the service layer depends only on abstractions (repository interfaces), not on the database details. This makes testing the service simpler since it does not know anything about the DB and thus does not require mocking the DB connection.

app/services/headerService.test.ts
```diff
@@ -0,0 +1,54 @@
+import { Locale } from 'next-intl';
+
+import { createHeaderService } from '~/services/headerService';
+
+jest.mock('~/db/connect', () => ({
```
Iarynovskyi: Here as well

app/shared/components/Footer/footer-social-media/social-media-icon/SocialMediaIcon.tsx
```diff
@@ -11,7 +12,7 @@ interface SocialMediaIconProps {
 
 const SocialMediaIcon = ({ icon, href }: SocialMediaIconProps) => {
   return (
-    <Link href={href} key={icon}>
+    <Link href={href} key={icon} target="_blank">
```
Iarynovskyi, line 15: Just curious, why did you change it here and change `import { Link } from '~/i18n/navigation'`  to `import { Link } from '@mui/material'` ?

app/di/container.ts
```diff
@@ -0,0 +1,11 @@
+import { createContainer } from 'awilix';
+
+import { registerRepositories } from './modules/repositories.module';
+import { registerServices } from './modules/services.module';
+
+export function createRequestContainer() {
```
Iarynovskyi, line 9: I think we may encounter a problem where multiple containers will be created when requests are made in different components, as I see it

`const { supportButtonLink } = await createRequestContainer().resolve('headerService').getHeaderData(locale);`

so make DI-container as a singleton

app/shared/components/Footer/Footer.consts.ts
Iarynovskyi, line 1: Do not forget to also remove the phrases from the translations files that now come from DB

app/models/navigation/navigation.ts
```diff
@@ -0,0 +1,28 @@
...
+    en: { type: String, required: true }
+  },
+  { _id: false }
+);
+
+const linkSchema = new mongoose.Schema(
+  {
+    label: { type: translatedFieldSchema, required: true },
+    href: { type: String, required: true },
+    visibility: { type: String, enum: ['true', 'false'], default: 'true' }
```
bandvov: Why visibility is string? Shouldn't it be boolean?

app/services/footerService.ts
bandvov: It should be moved to repository layer

app/shared/components/Footer/footer-social-media/social-media-icon/SocialMediaIcon.tsx
Mike-Popovych: I changed it because the links in this component point to external websites (like Instagram or Facebook), so we don’t need to use Next.js internal navigation.

app/repositories/foundation-info/foundationInfo.repository.ts
Mike-Popovych: I’ll remove the contactButtonLink — I just re-read the user story, and the button is supposed to open a modal window, not navigate anywhere.

app/models/navigation/navigation.ts
Mike-Popovych: Yes, I’ve fixed it

app/repositories/foundation-info/foundationInfo.repository.ts
Iarynovskyi: OK, i see that you delete it now from `faundationContactInfo` but here you forgot to delete it, so please delete it here too

app/repositories/navigation/navigation.ts
Iarynovskyi, line 1: Keep the same style of naming please
Or `some.repository.ts` like you wrote in `/fundation-info` folder
Or without `.repository.` suffix like you wrote in this file

app/types/types/dto/foundationInfo.dto.ts
Iarynovskyi, line 1: **This comment for all `dto`, `repository` and `services` folders in `types` folder**

This types looks like domain layer in DDD patern and maybe we can move them to some separate `domain` folder outside from this `types` ?
And i think that all logic with db we can move to `infrastructure` folder ( repositories, db-connection, models and maybe validators, because now i see that we use Zod only for db )

In this way, it seems to me that a project based on this pattern will be better structured

app/domain/dto/foundationInfo.dto.ts
```diff
@@ -0,0 +1,24 @@
+export type ContactInfoData = {
```
Iarynovskyi:
```suggestion
export type ContactInfoDTO = {
```
I think it would be more understandable  
If we write SomeDTO naming for all DTOs

app/infrastructure/models/navigation/navigation.ts
```diff
@@ -0,0 +1,28 @@
+import mongoose from 'mongoose';
+
+const translatedFieldSchema = new mongoose.Schema(
+  {
+    uk: { type: String, required: true },
+    en: { type: String, required: true }
+  },
+  { _id: false }
+);
```
Iarynovskyi: Move this `translatedFieldSchema` to some separate file, because you use it in many models, not only here

app/infrastructure/models/foundation-info/foundationInfoPublic.ts
```diff
@@ -0,0 +1,27 @@
...
+        copyright: {
+          uk: { type: String, required: true },
+          en: { type: String, required: true }
+        },
+        links: [
+          {
+            label: {
+              uk: { type: String, required: true },
+              en: { type: String, required: true }
+            },
```
Iarynovskyi: For example here

app/infrastructure/repositories/foundation-info/foundationInfo.repository.ts
```diff
@@ -0,0 +1,65 @@
...
+    const parsed = publicInfoSchema.parse(result);
+
+    return {
+      copyright: parsed.copyright[locale],
+      links:
+        parsed.links?.map((link) => ({
+          label: link.label[locale],
+          href: link.href
+        })) ?? []
+    };
```
Iarynovskyi: Maybe not good time for this, but here repository level know about user locale and if i correct know its bad
Only service or API layer can know about it
So maybe we can return here only parsed data and than in service layer use another zod schema that will be used like you here return data here ?
Знімок екрана 2025-07-02 о 17 43 14

So what you think about it ? Because its only my spontaneous idea

Mike-Popovych: Good idea, I’ll fix it.

Iarynovskyi (approved):
Awesome job!

danikua (approved):

bandvov (approved):

PR #221, opened 2025-06-09
Implemented ImageWithCaption component

### LINKED ISSUE #197: (SP:2) Implement ImageWithCaption component
Implement ImageWithCaption component. It consists of image, caption and border.

- [ ] Implement ImageWithCaption component
- [ ] Cover component with unit tests

**It should look like this:**
Image

**Or like this:**
Image

VKormylo:
Implemented ImageWithCaption component. This is reusable component for images with caption. You can specify sizes for different breakpoints for both image and border.

**Component takes these props:**

- src: string - src of image
- alt: string - alt for image
- caption: string - caption text
- sizes: object - responsive values for width and height
- border: object - placement of border and it's sizes
- containerSx?: sx - sx styles for container
- captionSx?: sx - sx styles for caption

**Completed tasks:**

- [x] Implemented ImageWithCaption component
- [x] Covered component with unit tests
- [x] Ensured it is responsive according to figma mockups

**Demonstration of how it looks:**

https://github.com/user-attachments/assets/9cd486db-d6d1-46bb-a200-0a679648b291

**Used code:**
image

**Results of tests:**
image
image

### DIFF
--- a/app/lib/utils/generateSizesAttribute.ts
+++ b/app/lib/utils/generateSizesAttribute.ts
@@ -7,7 +7,7 @@ type BreakpointKey = keyof typeof theme.breakpoints.values;
 export function generateSizesAttribute(sizes: Pick<ElementSizes, 'width'>): string {
   const breakpoints = theme.breakpoints.values;
 
-  const entries = Object.entries(sizes.width || {})
+  const entries = Object.entries(sizes.width)
     .filter(([bp]) => bp in breakpoints)
     .sort((a, b) => breakpoints[a[0] as BreakpointKey] - breakpoints[b[0] as BreakpointKey])
     .map(([bp, val], index, arr) => {
--- a/app/lib/utils/generateSizesAttribute.ts
+++ b/app/lib/utils/generateSizesAttribute.ts
@@ -2,17 +2,19 @@ import { theme } from '~/ds-components/theme/Theme';
 
 import { ElementSizes } from '~/types/types/common.types';
 
+type BreakpointKey = keyof typeof theme.breakpoints.values;
+
 export function generateSizesAttribute(sizes: Pick<ElementSizes, 'width'>): string {
   const breakpoints = theme.breakpoints.values;
 
-  const entries = Object.entries(sizes.width)
+  const entries = Object.entries(sizes.width || {})
     .filter(([bp]) => bp in breakpoints)
-    .sort((a, b) => breakpoints[a[0] as keyof typeof breakpoints] - breakpoints[b[0] as keyof typeof breakpoints])
+    .sort((a, b) => breakpoints[a[0] as BreakpointKey] - breakpoints[b[0] as BreakpointKey])
     .map(([bp, val], index, arr) => {
-      const max = breakpoints[bp as keyof typeof breakpoints];
+      const max = breakpoints[bp as BreakpointKey];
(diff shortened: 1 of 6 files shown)


uliaescha (approved):

app/shared/components/image-with-caption/ImageWithCaption.tsx
```diff
@@ -0,0 +1,47 @@
...
+  src,
+  alt,
+  sizes,
+  caption,
+  border,
+  containerSx = {},
+  captionSx = {}
+}) => {
+  const sizesAttibute = generateSizesAttribute(sizes);
+
```
danikua: you missed "r" in `sizesAttibute` and replace it below

Iarynovskyi (approved):

app/lib/utils/generateSizesAttribute.ts
```diff
@@ -0,0 +1,18 @@
...
+  const breakpoints = theme.breakpoints.values;
+
+  const entries = Object.entries(sizes.width)
+    .filter(([bp]) => bp in breakpoints)
+    .sort((a, b) => breakpoints[a[0] as keyof typeof breakpoints] - breakpoints[b[0] as keyof typeof breakpoints])
+    .map(([bp, val], index, arr) => {
+      const max = breakpoints[bp as keyof typeof breakpoints];
+      const isLast = index === arr.length - 1;
+      return isLast ? `${val}px` : `(max-width: ${max}px) ${val}px`;
+    });
```
danikua: The function looks good overall, but I have a few suggestions to improve readability and correctness:

1. Use a type alias for breakpoint keys to avoid repetitive casting and make the code cleaner. You can use it like this:

```ts
breakpoints[bp as BreakpointKey]
```

2. Sort breakpoints in descending order instead of ascending. The HTML `sizes` attribute expects media conditions from largest to smallest, with the last value as the default fallback. So the sorting should be:

```ts
.sort((a, b) => breakpoints[b[0] as BreakpointKey] - breakpoints[a[0] as BreakpointKey])
```

3. Add a fallback return value, for example `'100vw'`, if the input is empty to avoid returning an empty string.

app/shared/components/image-with-caption/ImageWithCaption.tsx
VKormylo: Thanks, gonna fix that.

app/lib/utils/generateSizesAttribute.ts
VKormylo: About type alias I agree. Also I will add fallback value, thanks. 
The HTML sizes attribute expects media conditions from largest to smallest, with the last value as the default fallback. - in the docs i saw that they should be from lowest to highest instead, because it matches first true condition and if we have something like this `sizes="(max-width: 1920px) 50vw, (max-width: 1024px) 75vw, 100vw"` it will match only first one always.

danikua: OK, got it

VKormylo: Done.

app/shared/components/image-with-caption/ImageWithCaption.tsx
VKormylo: Done.

danikua (approved):

PR #289, opened 2025-06-20
Design System Scrollable container

### LINKED ISSUE #259: (SP:1) Scrollable Container
Description:
Implement reusable custom scrollable container 
Example:
Image

yur4uwe:
## Summary of issue

Create Scrollable container component

## Summary of change

Created scrollable element with custom scroller styles

## Testing approach

Just test for some basic things as logic isn't that complex

## CHECK LIST

- [x] 100% test coverage
- [x] scroller works 👍 (trust me bro)
- [x] browser specific


app/shared/components/design-system/all-components/scrollable/Scrollable.tsx
```diff
@@ -0,0 +1,41 @@
...
+  '&::-webkit-scrollbar-thumb': {
+    backgroundColor: mainHexPallete.blue[400],
+    borderRadius: '16px',
+    border: '3px solid transparent',
+    backgroundClip: 'content-box',
+    '&:hover': {
+      backgroundColor: mainHexPallete.black
+    }
+  }
+});
```
danikua: Should we replace `width: min-content` with `100%` or something like this for better responsiveness? Using `min-content` for width might cause unexpected layout behavior if the content is wider than expected. The container may not stretch to fill the available space or might overflow. 

And please, write styles in a separate file.

app/shared/components/design-system/all-components/scrollable/Scrollable.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface ScrollableProps extends BoxProps {
+  children: React.ReactNode;
```
luvthenika, line 8: i think this class extension is redudant as BoxProps already includes children as it is anotomy is a simple div with MuiBox-root class 
https://mui.com/material-ui/react-box/#anatomy

app/shared/components/design-system/all-components/scrollable/Scrollable.tsx
```diff
@@ -0,0 +1,41 @@
+'use client';
+
+import { Box, BoxProps, styled } from '@mui/material';
+import React from 'react';
+
+import { mainHexPallete } from '../theme/colors';
+
+const NativeScrollableContent = styled(Box)({
```
luvthenika: in order to follow agreed theme design architecture , i suggest moving style defnitions into the Theme.ts and create custom variant with these styles

app/shared/components/design-system/all-components/scrollable/Scrollable.tsx
yur4uwe: done ;P

danikua (approved):

app/shared/components/design-system/all-components/scrollable/Scrollable.tsx
yur4uwe: The thing is that BoxProps do include `children` prop but its optional, but here, as `Scrollable` is a wrapper it must have children, that's why i added that prop

app/shared/components/design-system/all-components/scrollable/Scrollable.tsx
yur4uwe: Mui doesn't allow overwriting `Box` component styles at all, and there is no component with similar functionality. So my only bet is sx prop, global css or custom styled element

Iarynovskyi (approved):

IrynaKhylchuk (approved):

PR #291, opened 2025-06-20
Implemented emotion cache

### LINKED ISSUE #199: (SP:3) Set Up Emotion Cache for correct MUI and Next.js integration
Ensure proper styling and hydration when using Material UI (MUI) with Emotion in a Next.js project. Prevent SSR/CSR mismatch errors by setting up a shared Emotion cache and wrapping the app with CacheProvider.

Currently there are hydration errors due to styles mismatch.

Implementing EmotionProvider will fix hydration errors and allow usage of hooks from MUI (useTheme).

- [ ] Create createEmotionCache.ts to configure a custom Emotion cache.
- [ ] Implement EmotionProvider.tsx to wrap the app with CacheProvider using that cache.
- [ ] Wrap the root layout (layout.tsx) with EmotionProvider above any MUI components or context providers.
- [ ] Ensure Emotion SSR compatibility by verifying correct style injection and preventing hydration mismatch.

VKormylo:
Implemented emotion cache. Added EmotionProvider that ensures that styles on both client and server is same. There is no hydration errors now.

**No hydration errors:**
image

**Result of the tests:**
image


IrynaKhylchuk (approved):

app/[lang]/layout.tsx
```diff
@@ -52,47 +53,52 @@ export default async function RootLayout({ children, params }: RootLayoutParams)
...
+                    [theme.breakpoints.down('md')]: {
+                      gridTemplateColumns: 'repeat(8, 1fr)',
+                      paddingLeft: '24px',
+                      paddingRight: '24px',
+                      gap: '20px'
+                    },
+                    [theme.breakpoints.down('sm')]: {
+                      gridTemplateColumns: 'repeat(4, 1fr)',
+                      gap: '16px'
+                    }
```
Comment (danikua) [inline app/[lang]/layout.tsx:87]: It might be better to write the styles in a separate file to follow project pattern

app/[lang]/layout.tsx
```diff
@@ -52,47 +53,52 @@ export default async function RootLayout({ children, params }: RootLayoutParams)
...
+                  }}
+                >
+                  <Box sx={{ gridColumn: '1 / -1' }}>
+                    <Header />
+                  </Box>
+                  <Box sx={{ display: 'grid', gridTemplateColumns: 'subgrid', gridColumn: '1 / -1' }}>{children}</Box>
+                  <Box sx={{ gridColumn: '1 / -1' }}>
+                    <Footer />
```
Comment (danikua) [inline app/[lang]/layout.tsx:95]: same

app/shared/components/emotion-provider/EmotionProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const flush = () => {
+      const prevInserted = inserted;
+      inserted = [];
+      return prevInserted;
+    };
+
+    return { cache, flush };
+  });
+
```
danikua: I think it would be better for readability:

```
const [emotionState] = useState(() => {
  ...
});
const { cache, flush } = emotionState;
```

app/[lang]/layout.tsx
Comment (VKormylo) [reply app/[lang]/layout.tsx:87]: I don't think i should change it in my PR. I think it's better to refactor this in PR with grid fixing.

app/shared/components/emotion-provider/EmotionProvider.tsx
VKormylo: Yeah, i agree, thanks.

app/[lang]/layout.tsx
Comment (VKormylo) [reply app/[lang]/layout.tsx:95]: Same response here as well.

app/[lang]/layout.tsx
Comment (VKormylo) [reply app/[lang]/layout.tsx:87]: I just wrapped it into provider, that's why it shows it as changed lines.

Comment (danikua) [reply app/[lang]/layout.tsx:87]: OK, then i think it will be a good idea to create another issue to fix project styles and separate them

danikua (approved):

Iarynovskyi (approved):

PR #306, opened 2025-06-25
Implement year filtering for tables

### LINKED ISSUE #271: (SP:1) Create year filtering for tables
Implement filtering by years 

AC:
- should correspond to design
- should be able to display chosen years

Image

IrynaKhylchuk:
Implemented **year filtering** for tables.

Key changes: 
- added new `YearFiltering` component;
- applied styles consistent with the design system;
- added unit tests for `YearFiltering`;
- integrated **i18n** translations for year-related labels and error messages;
- refactored `ColoredSvg` component;
- refactored `Slider` component;


app/shared/components/design-system/all-components/filtering/year-filtering/YearFiltering.tsx
Mike-Popovych, line 1: Would it be ok to refactor this component to remove the internal state and make it fully controlled instead?
- It will always rely on the `value` prop as the single source of truth.
- All updates will happen via `onChange`, so the parent manages the state.

Also, would it make sense to generalize this component so it’s not tied specifically to years?

app/shared/components/design-system/all-components/filtering/year-filtering/YearFiltering.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+  default: () => <svg data-testid="trash-icon" />
+}));
+
+const minYear = 1930;
+const maxYear = 2020;
+
+describe('YearFiltering', () => {
+  beforeEach(() => {
+    render(<YearFiltering minYear={minYear} maxYear={maxYear} />);
+  });
```
Kryzhanivsky, line 31: Missing spacing between code blocks.

Kryzhanivsky (approved):

PR #369, opened 2025-07-14
fix: fixed an issue with the opus icon display

### LINKED ISSUE #359: (SP: 0.5) Bug with open table opus button
When we go to page /artistry we not see open opus row icon
And see it only when it opened
Need to fix problem

Go to http://localhost:3000/en/artistry
And you can see problem

**Now:**
https://github.com/user-attachments/assets/e6d8a029-fa78-407f-8f6d-1c34df6f43da

**Need to be:**
https://github.com/user-attachments/assets/42859974-175a-46f1-9a2b-0a34ea7556c0

Kryzhanivsky:
## Summary of change

- Fixed an issue with the opus open row icon being displayed
- Tests adapted to new functionality
- Edited the settings of some svg icons

Screenshot:
image

### DIFF
--- a/app/shared/components/enhanced-table/collapsible-row/CollapsibleRow.tsx
+++ b/app/shared/components/enhanced-table/collapsible-row/CollapsibleRow.tsx
@@ -4,14 +4,18 @@ import { Box, TableCell, TableRow } from '@mui/material';
 import { type ColumnDef, getCoreRowModel, useReactTable } from '@tanstack/react-table';
 import React from 'react';
 
+import { Svg } from '~/components/colored-svg/ColoredSvg';
 import { IconButton } from '~/ds-components/icon-button/IconButton';
+import { mainHexPallete } from '~/ds-components/theme/colors';
 
 import { CollapsibleDataRow } from './CollapsibleDataRow';
 import { collapsibleRowStyles as styles } from './CollapsibleRow.styles';
 import { IconButtonColorVariant } from '~/types/enums/common.enums';
 import type { CollapsibleGroupColumnMeta, RowData } from '~/types/types/enhancedTable';
 
-import { SvgImage } from '~/shared/components/svg-image/SvgImage';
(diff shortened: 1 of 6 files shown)


Iarynovskyi (approved):

app/shared/components/enhanced-table/collapsible-row/CollapsibleRow.tsx
```diff
@@ -47,11 +51,12 @@ export function CollapsibleRow<T extends RowData>({
...
-                    <SvgImage
-                      src={collapsed ? '/icons/chevron-down.svg' : '/icons/chevron-right.svg'}
+                    <Svg
+                      Component={collapsed ? chevronDown : chevronRight}
+                      color={mainHexPallete.brown['700']}
                       alt="toggle"
-                      width={24}
-                      height={24}
+                      width="24px"
+                      height="24px"
```
oleg191006: in our Svg component the height and width are 24 by default so I think you can leave them out here

danikua (approved):

app/shared/components/enhanced-table/collapsible-row/CollapsibleRow.tsx
Kryzhanivsky: In the future, there may be changes that will change the icon sizes. For example, replacing the default sizes or the entire icon. Wouldn't it be better to leave the icon sizes specified here? What do you think?

oleg191006: > In the future, there may be changes that will change the icon sizes. For example, replacing the default sizes or the entire icon. Wouldn't it be better to leave the icon sizes specified here? What do you think?

Well, maybe, but in my opinion, this is just extra code for now, and if necessary, it will be no problem to add these props back

Kryzhanivsky: code updated

oleg191006 (approved):

PR #429, opened 2025-07-31
Header data fetch

### LINKED ISSUE #417: (SP: 1) Fetch and Render Header Data
Create an API route /api/header that returns header data from the database using headerService, and integrate it into the Header component via useQuery.

 Requirements
 - Create app/api/header/route.ts that:

 - Uses createRequestContainer() and resolves headerService

 - Calls getHeaderData(locale) with getLocale() from next-intl/server

 - Returns navigation and supportButtonLink as JSON

 - In the Header component:

 - Use useQuery directly to fetch data from /api/header

Pass navigation to <NavigationBar /> and supportButtonLink to <RightActionsPanel />

### LINKED ISSUE #427: (SP: 1) Implement utility for generating full URLs with dynamic and query parameters
**Description:**  
Create a utility function that constructs a complete URL from a pathname template with optional dynamic segments and query parameters.

---

#### Requirements:

- Support dynamic route segments like `/path/[id]` that are replaced with actual values.
- Accept optional query parameters that may include:
  - primitives: `string`, `number`, `boolean`
  - arrays of primitives
  - flat objects with primitive values
- Skip `null` and `undefined` values.
- Ensure correct URL encoding of query parameters.
- Support nested query structures using the format:  
  - Arrays: `param[0]=value1&param[1]=value2`  
  - Objects: `param[key]=value`

---

### LINKED ISSUE #439: (SP:0.5) Fix missing active state in ButtonGroup for dynamic nav
## Description
While testing and rendering the DesktopNav component, the active button highlighting (indicator and color) worked correctly when using hardcoded navigation items but failed when loading navigation data from the backend (e.g., via navLabels fetched from an API or mocked dynamically).

## Problem

The ButtonGroup correctly highlighted the active button when buttons were hardcoded directly in the component or test.

However, when navLabels were fetched from the backend or passed as mock props (e.g., JSON), the indicator did not appear under the active item, and the text color did not reflect the active state.

Root cause: getBoundingClientRect returned 0 width/left because the real DOM layout isn't available in the test or render environment unless these values are mocked.

The component relies on these DOM measurements to position the indicator under the active item.

Mike-Popovych:
## Description

Integrated dynamic navigation data into the header:

- Implemented `/api/header` API route that returns localized navigation and support button link.
- Created `headerClientService` to fetch header data on the client side.
- Updated the `Header` component to fetch and render data using a custom `useQuery` hook.
- Replaced hardcoded navigation values in `DesktopNav` with dynamic data from the API.
- The API already returns localized values, so `locale` handling in the component is no longer necessary.

## Type of change

- [x] New feature  

## How to test

1. Run the project locally.
2. Verify that `Header` and `DesktopNav` render correct navigation based on API data.
3. Change locale (`/uk`, `/en`, etc.) and ensure correct localized labels are shown.
4. Check that the support button link is correctly rendered.

## Video / Screenshots

https://github.com/user-attachments/assets/7b0b37e3-de67-4acf-9ba4-f838257a311a

## Checklist

- [x] Tests pass successfully  
- [x] No Sonar issues   


oleg191006:
Where did you find this header option? I just don't see it in figma at all

Kryzhanivsky (approved):

app/api/header/route.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function GET(req: NextRequest) {
+  const locale = req.nextUrl.searchParams.get('locale') || 'uk';
+
+  try {
+    const container = createRequestContainer();
+    const navigationService = container.resolve('navigationService');
+    const foundationInfoService = container.resolve('foundationInfoService');
+
+    const headerService = createHeaderService({ navigationService, foundationInfoService });
```
Iarynovskyi:
```suggestion
    const container = createRequestContainer();
    const headerService = container.resolve('headerService');
```
You can resolve `headerService` with its cores services inside with this one line of code, i tried it and its work
If we use a DI-container, then we need to follow the rules of layered architecture and not import the service directly

Iarynovskyi (approved):

danikua (approved):

PR #441, opened 2025-08-01
(SP: ) Create preview API endpoint

### LINKED ISSUE #407: (SP:3 ) Create preview API endpoint
## Description:
Implement an API endpoint that accepts parameters like access token, page slug, language, and draft ID (optional)
The endpoint verifies the token, enables preview mode (draftMode()), sets necessary cookies, and redirects to the requested page

### Steps:

- Receive parameters: slug, previewToken, lang, draftId (optional)
- Validate the preview token securely
- Call draftMode().enable() to activate preview mode
- Return a redirect response to the requested page URL
- Set headers to prevent caching on preview pages

oleg191006:
## Description

- Added API route to handle preview mode

- Implemented utility to decode JWT tokens

- Added helper to extract Bearer token from request headers

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test
you can see t in the video, you also need to add JWT_SECRET and PREVIEW_SECRET to the .env.local file
## Checklist

https://github.com/user-attachments/assets/84a087fd-c310-46c0-ac1b-ab67c832a6af

- [x] Tests pass successfully
- [x] No Sonar issues

---


app/api/preview/route.ts
```diff
@@ -0,0 +1,55 @@
...
+  const slug = searchParams.get('slug');
+  const token = searchParams.get('previewToken');
+  const lang = searchParams.get('lang');
+  const draftId = searchParams.get('draftId');
+
+  if (!slug || !token || !lang) {
+    return errorResponse([errors.MISSING_PARAMETERS], 400);
+  }
+
+  const VALID_TOKEN = process.env.PREVIEW_SECRET;
```
Iarynovskyi: I think this `process.env` would be better to move to config file and get it from there 
And do same thing with  secret token in `app/lib/utils/verifyToken.ts` file too

oleg191006: I agree, thank you

Iarynovskyi (approved):

danikua (approved):

app/api/preview/route.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (!tokenFromHeader) {
+    return errorResponse([errors.MISSING_AUTH_HEADER], 401);
+  }
+
+  const user = verifyToken(tokenFromHeader);
+  if (!user) {
+    return errorResponse([errors.INVALID_TOKEN], 401);
+  }
+
+  if (user.type !== 'admin' && user.type !== 'superadmin') {
```
bandvov, line 27: should it be logical OR to check if user is either admin OR supeadmin?
```react
if (user.type !== 'admin' || user.type !== 'superadmin') {
...
}
```

oleg191006: well, in this case we need to check that the user is not an admin and not a superadmin, so we need to use the logical operator AND

app/lib/utils/verifyToken.ts
```diff
@@ -0,0 +1,15 @@
+import jwt from 'jsonwebtoken';
+
+import type { AdminTokenPayload } from '~/types/types/admin.types';
```
bandvov: Maybe it make sence rename to `AuthTokenPayload` because now we have only admins but it may change in the future

app/lib/utils/verifyToken.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { jwtSecret } from '~/config';
+
+export function verifyToken(token: string): AdminTokenPayload | null {
```
bandvov: Do we want to use different types of tokens(auth,preview)? is so consider to give more specific name to know what you want to verify. `verifyAuthToken` for example. or keep old name but update function to accept secret as parameter

app/lib/utils/verifyToken.ts
oleg191006: yes indeed, already changed

app/lib/utils/verifyToken.ts
oleg191006: done

Kryzhanivsky (approved):

luvthenika (approved):

PR #495, opened 2025-08-13
(SP: 0,5) Bug with table sorting 

### LINKED ISSUE #468: (SP: 0,5) Bug with table sorting
The table sorting worked till a certain moment. After some changes the table sorting stopped working.

Image

luvthenika:
## Description
Enabled and made sorting work
https://github.com/user-attachments/assets/404dcbc6-0bd6-4475-bec3-bb7c1a9970a7


uliaescha (commented):
In our design, there are no arrows near the column names, so I don't think we need to enable sorting in that way
image

luvthenika:
composition tables doesnt need sorting according to the design

Issue #481, opened 2025-08-11
(SP: 1) Create a custom Backdrop component based on Mui Modal.


Kryzhanivsky:
## Description

- [ ] - Create a component based on the Modal component from the Material UI library.
- [ ] - Write tests for the created component

This component should support:

- [ ] color #190D03
- [ ] transparency 25%
- [ ] blur 4
- [ ] **Optional:** Support positioning via props

⚠️**Attention:** This component is just a background. The white component and its fill are processed in other tasks.

## Example:

<img width="1136" height="675" alt="Image" src="https://github.com/user-attachments/assets/c19c4ca0-d0f8-41fa-9dfa-29ab3f0bec25" />

PR #519, opened 2025-08-19
Add custom paper component

### LINKED ISSUE #482: (SP: 2) Create a custom Paper component based on Mui Paper.
## Description

- [ ] Create a component based on the Modal component from the Material UI library.
- [ ] Write tests for the created component

This component should: 

- [ ] Be universal (not only for modal windows)
- [ ] Have the ability to change its own styles
- [ ] Change the height dynamically

⚠️**Attention:** The component does not provide predefined elements inside. All internal elements are passed through the children prop.

## Examples

Image

Image

ssashayurchenko:
## Description

Implemented a custom PaperComponent based on MUI Paper.
I also removed redundant styles from `ContactForm` since they are now provided by the `PaperComponent` wrapper.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test
Wrap your content with the `PaperComponent`.
- Use the `sx` prop to customize styles for the `Paper` wrapper.
- Use the `childrenSx` prop to customize styles for the content inside the `Paper`.
- Use the `elevation` prop to adjust the `shadow` (default is `0`).

```
<PaperComponent 
  sx={{ backgroundColor: 'white', width: 500, height: 700 }} 
  childrenSx={{ padding: '40px' }} 
>
  <ContactForm />
</PaperComponent>
```

## Video / Screenshots
Знімок екрана 2025-08-19 о 18 52 48

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---


app/shared/components/paper-component/PaperComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+  sx?: object;
+}>;
+
+export default function PaperComponent({ children, isModal, onClose, sx }: PaperComponentProps) {
+  return (
+    <Paper
+      sx={{
+        ...styles.container,
+        ...(isModal ? styles.modal : styles.block),
+        ...sx
```
Kryzhanivsky: There is another task #481 to create the modal window logic. This component should not contain this functionality.

app/shared/components/paper-component/PaperComponent.styles.ts
```diff
@@ -0,0 +1,42 @@
+import { mainHexPallete } from '../design-system/all-components/theme/colors';
+
+export const styles = {
+  container: {
+    backgroundColor: mainHexPallete.white,
+    padding: {
+      xs: '40px 24px',
+      sm: '64px 32px',
+      md: '76px 53px',
+      lg: '90px 80px'
```
Kryzhanivsky: This component should be completely externally controlled. Let's discuss this at the meeting.

app/shared/components/paper-component/PaperComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+      }}
+    >
+      {isModal && onClose && (
+        <IconButton onClick={onClose} sx={styles.icon}>
+          <SvgImage src="/icons/x.svg" alt="closing modal" width={40} height={40} />
+        </IconButton>
```
Kryzhanivsky: I think this logic should be taken outside the component and removed from here.

app/shared/components/paper-component/PaperComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+      )}
+      <div style={{ transform: 'skewY(2deg)', width: '100%' }}>{children}</div>
+    </Paper>
```
Kryzhanivsky, line 21: Can we use Box and remove the inline styles from here?

app/shared/components/paper-component/PaperComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { IconButton } from '../design-system/all-components/icon-button/IconButton';
+import { SvgImage } from '../svg-image/SvgImage';
+import { styles } from './PaperComponent.styles';
+
+type PaperComponentProps = Readonly<{
+  children: ReactNode;
+  isModal: boolean;
+  onClose?: () => void;
+  sx?: object;
```
Kryzhanivsky: Use SxProps<Theme> instead of object

app/shared/components/forms/contact-form/ContactForm.tsx
Kryzhanivsky, line 1: The ContactForm component is not relevant to this task. If you need to change something, you should do it in a separate task.

Kryzhanivsky (changes requested):

Mike-Popovych:
image

According to the design, we should not have a large bottom margin or a box shadow.

My testcase
image

app/shared/components/paper-component/PaperComponent.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+    width: '100%',
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: 'center',
+    alignItems: 'center'
```
Mike-Popovych:
```suggestion
    alignItems: 'center',
     boxShadow: 'none'
```
there is no box shadow in our design, so we can remove it

Mike-Popovych (approved):

app/shared/components/paper-component/PaperComponent.tsx
ssashayurchenko: > #481

As stated in my task, the component should support being used either as a modal window or as a simple block. The isModal prop simply controls whether a close icon is rendered (if it’s a modal window) or not (if it isn’t). Therefore, I haven’t implemented any logic from task [#481](https://github.com/orgs/Liatoshynsky-Foundation/projects/1/views/1?filterQuery=481&pane=issue&itemId=123754837&issue=Liatoshynsky-Foundation%7Clf-client%7C481), such as using the Modal component from MUI, adding blur effects etc.

app/shared/components/forms/contact-form/ContactForm.tsx
ssashayurchenko: > The ContactForm component is not relevant to this task. If you need to change something, you should do it in a separate task.

The changes I made in ContactForm are part of my task. The wrapper has a fixed width and controls the background color, which makes it hard to adjust the component. I also removed one padding (pl: '12px') since all other paddings are handled by the Paper wrapper.

app/shared/components/paper-component/PaperComponent.styles.ts
ssashayurchenko: > there is no box shadow in our design, so we can remove it

Ok, I think it might be better to use the elevation prop of the Paper component to remove the shadow.

luvthenika (approved):

app/shared/components/paper-component/PaperComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { sxToArray } from '~/lib/utils/sxToArray';
+
+type PaperComponentProps = Readonly<{
+  children: ReactNode;
+  padding: ResponsiveStyle;
+  backgroundColor: (typeof mainHexPallete)[keyof typeof mainHexPallete] | string;
+  sx?: SxProps<Theme>;
+}>;
+
+export default function PaperComponent({ children, sx, padding, backgroundColor }: PaperComponentProps) {
```
Kryzhanivsky: Usually, background color and padding are passed through styles, so separate props are redundant. Moreover, they are mandatory, which obliges us to use them in every call.

In my opinion, it would be better to abandon our own type altogether and use the props of the Paper component from the Mui library:
```suggestion
interface PaperComponentProps extends PaperProps {
  childrenSx: SxProps<Theme>;
}

export default function PaperComponent({ children, sx, childrenSx, elevation = 0, ...props }: PaperComponentProps ) {
```

Then we can simply pass props to the Paper component, but add custom styles.

This way, we will get rid of unnecessary types and inline styles.

Kryzhanivsky (changes requested):

app/shared/components/paper-component/PaperComponent.tsx
ssashayurchenko: > Usually, background color and padding are passed through styles, so separate props are redundant. Moreover, they are mandatory, which obliges us to use them in every call.
> 
> In my opinion, it would be better to abandon our own type altogether and use the props of the Paper component from the Mui library:
> 
> Then we can simply pass props to the Paper component, but add custom styles.
> 
> This way, we will get rid of unnecessary types and inline styles.

padding and backgroundColor are always required by design, which keeps the component consistent with the design system

Kryzhanivsky: Yes, but you still pass them inline styles later. It's better to pass them in the style object right away. Also, there might be a need for a default style somewhere, then you don't need to pass them at all, and with the current implementation that's impossible.

Kryzhanivsky (approved):

uliaescha (approved):

PR #719, opened 2025-10-21
Move the page Layout outside the Next.js structure.

### LINKED ISSUE #710: (SP: ) Move the page Layout outside the Next.js structure.
## Description
It is necessary to move the layout logic for pages outside the Next.js structure. This will reduce code duplication and file nesting in the project. It will also expand the possibility of adding new page templates in the future.

Image

Image

## Steps
- [ ] Move the logic for Layout into a separate component
- [ ] Remove unnecessary nesting by moving page folders to the [lang] directory
- [ ] If necessary, correct tests according to new changes

## Expected Result

The project is launched and works as before the changes.

Kryzhanivsky:
## Description

Added new Layouts that are not included in the standard Next.js folder structure.
Now the main Layout component adds a header with a footer and sets the base size of the container. Positioning and padding on the page will now be handled by custom Layouts that are not included in the Next.js page directory structure.

This will reduce the nesting of directories with pages and code duplication. It will also allow for more flexible management of page templates in the future.

## How to use

Now the page component can be wrapped in a custom Layout:

```
return (
    <MainLayout>
        <SomeComponents />
    </MainLayout>
  );
```

```
return (
    <MainLayout withLines>
        <SomeComponents />
    </MainLayout>
  );
```

```
  return (
    <ColoredLayout>
      <SomeComponents />
    </ColoredLayout>
  );
```

## Video / Screenshots

image

image

## Checklist

- [ ] Tests pass successfully
- [ ] No Sonar issues

---


app/shared/layouts/colored-layout/ColoredLayout.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <Box sx={[styles.container(color), ...(Array.isArray(sx) ? sx : [sx])]} data-testid="colored-layout" {...props}>
+      <MainLayout
+        sx={[styles.childrenBox, ...(Array.isArray(mainLayoutSx) ? mainLayoutSx : [mainLayoutSx])]}
+        {...otherProps}
+      >
+        {children}
+      </MainLayout>
+    </Box>
+  );
+};
```
yur4uwe, line 21: I don't think we should expose any styles of the background color box aside from `backgroundColor`. This way we can simplify `ColoredLayout` and make its props extend `MainLayoutProps` with optional color as prop. So we get someting like this:

```suggestion
export interface ColoredLayoutProps extends MainLayoutProps {
    color?: string;
}

const ColoredLayout: React.FC<ColoredLayoutProps> = ({
    color = '#F2EEE8',
    children,
    sx,
    ...props
}) => (
    <Box sx={{ backgroundColor: color }} data-testid="colored-layout">
        <MainLayout
            sx={[styles.childrenBox, ...(Array.isArray(sx) ? sx : [sx])]}
            {...props}
        >
            {children}
        </MainLayout>
    </Box>
);
```

Kryzhanivsky: I understand that this is not the best practice, but I wanted to provide maximum flexibility and code reuse. Also, the container also has its own specific styles in the ColoredLayout.styles.ts file. Your version looks better, but it is less flexible.

yur4uwe: Sorry, i missed the styles part. But we don't really need this much of the flexibility. It like to premature optimisation but with flexibility, I think for now its better to make more understandable interface, bacuase we have no idea what will happen in the future, and refactoring simple interface is much easier than a flexible one

Kryzhanivsky: I think we need a third party opinion.

yur4uwe: Agree

bandvov: Let's use Roman's approach(two separate independent layouts). As for me it is more flexible and will avoid using unexpected styles.

yur4uwe: Well there still would be 2 layouts, just  `ColoredLayout` will inherit from `MainLayout`. Background Box only reason to exist is to provide background color without being limited by width of main layout, we really don't need to modify its `sx`, everything could be done through `MainLayout` that requires style changes as all content will be inside. So my opinion still stands, we don't need to manipulate styles of background and simply let it exist with hard-coded styles

Kryzhanivsky: Now I understand what Yuriy was trying to convey, and I agree with it. I edited the layout, according to his wishes, but left the option to change the styles in the wrappers, if it is necessary at some point. However, I did not make both layouts independent, because, in my opinion, this violates the DRY principles, because the styles for the `MainLayout` will be duplicated in both components.

yur4uwe: Yes, this was almost the thing i was trying to say, excepti argued that we wouldn't need styles adjustments for background box but i can get behind that, you've got my approval :+1:

yur4uwe (approved):

Kryzhanivsky (commented):
I noticed that some of the indents were doubled and that I forgot to pass the props to change the line color, so I fixed that.

yur4uwe (approved):

yur4uwe:
> I noticed that some of the indents were doubled and that I forgot to pass the props to change the line color, so I fixed that.

I see...

Mike-Popovych (approved):

oleg191006 (approved):

PR #858, opened 2025-11-05
Refactor(filters): unify table filters into reusable component

### LINKED ISSUE #857: (SP: 1) Unify filter components
## Description

Refactor the existing filter components to improve consistency and avoid duplication.
Currently, MusicTableFilters and TableFilters contain overlapping logic for rendering filters and a "clear all" action.
We need to extract shared logic and ensure both components rely on the same implementation pattern.

## Expected Result

Shared, reusable, filter component with consistent behavior across tables, reduced code complexity, and improved maintainability.

qqwz0:
## Description

Refactored table filter components across MusicTableSection and WorkTableSection.
Standardized filter behavior, fixed ordering of filters and improved test coverage.
Rewrote tests to match the new data flow and filter interaction logic.

## Type of change

- [x] Refactoring / Tech debt

## How to test

1. Verify that filters work: genre / category / year / author.

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---


NatalyKrvch (approved):

stsvt (approved):

app/shared/components/design-system/all-components/table-filters/TableFilters.tsx
```diff
@@ -0,0 +1,68 @@
...
+  isAnyFilterActive: boolean;
+}
+
+export function TableFilters({ filters, onClearAllFilters, isAnyFilterActive }: Readonly<TableFiltersProps>) {
+  const t = useTranslations('filtering');
+
+  const orderedFilters = (() => {
+    const movable = filters.filter((f) => !f.isStatic);
+
+    const sortedMovable = [...movable].sort((a, b) => Number(b.isActive) - Number(a.isActive));
```
kolibri753: Tiny thought (I might be off): if isActive is ever undefined (and it's optional in the type), Number(undefined) becomes NaN, which can make the sort order a bit unstable. Maybe we coerce to booleans or even default `isActive = false` during destructuring

app/shared/components/design-system/all-components/table-filters/TableFilters.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import Delete from '~/public/icons/trash-2.svg';
+import TooltipCustom from '~/shared/components/design-system/all-components/tooltip/Tooltip';
+
+interface TableFilterItem {
+  element: React.ReactNode;
+  isActive?: boolean;
+  id?: string;
+  isStatic?: boolean;
+}
```
kolibri753, line 20: Maybe `id` should be required here? I think I saw it as a React key and in data-testid later. If a consumer forgets to pass it, React will warn and tests may break

app/shared/components/design-system/all-components/table-filters/TableFilters.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <Box sx={{ alignSelf: 'center' }} data-testid="TableFilters-clearButtonWrapper">
+            <TooltipCustom title={t('clearAll')} placement="top">
+              <IconButton
+                data-testid="TableFilters-clearButton"
+                type={IconButtonVariant.outlined}
+                variant={IconButtonColorVariant.Secondary}
+                size="medium"
+                onClick={onClearAllFilters}
+                sx={{ border: 'none', padding: 0 }}
+              >
```
kolibri753, line 60: Since this is an icon-only button, should we add an accessible name? `aria-label={t('clearAll')}`

kolibri753 (approved):
Looks good as is - dropping a couple of thoughts in case they help :)

app/shared/components/design-system/all-components/table-filters/TableFilters.tsx
qqwz0: Agree. Updated the sorting to use Boolean to avoid NaN and make ordering more stable

app/shared/components/design-system/all-components/table-filters/TableFilters.tsx
qqwz0: Valid point. Done!

app/shared/components/design-system/all-components/table-filters/TableFilters.tsx
qqwz0: Fixed! Now it is required

PR #865, opened 2025-11-06
Feature/782/phone number validate intl normalize to e164 include in request error messages tests

Taras-ep:
## Description

Added phone number masking, country code validation and normalization logic for the Contact Form.
The new implementation ensures that users see a formatted (masked) phone number while the backend receives a normalized version (only + and digits).

## Additionally:

 - Implemented the custom React hook useHandlePhoneInput to manage phone number masking, error state, and validation logic.
- Added the normalizePhoneNumberFromMask utility for converting a masked value into a backend-safe format.
- Created and passed unit tests for both the masking logic and normalization utility.
- Updated ContactForm to integrate the new masking and normalization logic.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test
1. Run the project:
2. npm run dev
3. Open the contact page with the form.
4. Type a phone number — the input should automatically apply a mask (+380 (63) 123-45-67, etc.).

Check that:
- You can only enter valid characters.
- Errors appear for invalid formats, too short or too long numbers.
- The format to store in database will be e164. Use normalizePhoneNumberFromMask to convert masked value.

Run tests to verify behavior:
1. npm run test
2. All new test cases should pass successfully.

## Checklist

- [ ] Tests pass successfully
- [ ] No Sonar issues

---


app/constants/phoneCountryCodes.ts
Iarynovskyi, line 1: I more than 100% sure that we can use some JS library for this instead of hardcoded data here

Iarynovskyi: I found this library `libphonenumber-js` that have validation for phone numbers and all codes for many countries 
So in my opinion it would be more better to use this or another one for this issue
_but its only my opinion so you can not listen me if you have a reason for this logic_

Taras-ep: Thanks for the suggestion! For this optional field we only need masking + simple normalization to E.164. Pulling a whole phone-number library adds bundle weight, extra runtime and supply-chain risk without giving us product value right now. Let’s keep the lightweight implementation; if we later need full multi-country parsing/formatting, we can add an adapter and revisit.

Also, if I'm not mistaken the autor of this library from rusia. (I wanted to use it because it is much lighter than the original Google one, but I didn't for the previous mentioned reason )

Taras-ep: I closed this PR because I had some issues with history. I opened new one, so you can leave your
 comments there and we will continue to discuss.

Here is link: https://github.com/Liatoshynsky-Foundation/lf-client/pull/880

PR #867, opened 2025-11-06
chore: refactor image with border component to support flexible dimen…

### LINKED ISSUE #861: (SP: 0.5)  ImageWithBorder dimensions refactor
## Description

Refactor the ImageWithBorder component to make width and height optional. The current fixed sizes limit responsiveness. The component should allow both explicit sizing and size control from the parent component.

## Expected Result
There is should be a possibility to specify width and height, and we should have the possibility to set the dimensions of the component from the parent container.

ruslansymonenko:
## Description

Refactored the ImageWithBorder component to allow both fixed dimensions and parent-controlled sizing, making it easier to use and improving site responsiveness.

## Type of change
- [x] Refactoring / Tech debt

## Screenshots

> Plese note! This is example screenshots, not website component!

### Component with fixed dimensions
imgDimensions
imgDimensionsExample

### Dimensions controlled by the parent
imgDimensions2
imgDimensionsExample2

### DIFF
--- a/app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
+++ b/app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
@@ -1,20 +1,20 @@
 import { mainHexPallete } from '../theme/colors';
 
 export const styles = {
-  container: (width: number, height: number) => ({
+  container: (width?: number, height?: number) => ({
     position: 'relative',
-    width: width,
-    height: height,
+    width: width ?? '100%',
+    height: height ?? '100%',
     gridColumn: '1 / -1'
   }),
-  border: (borderWidth: number, height: number) => ({
+  border: (borderWidth: number, height?: number) => ({
     display: 'block',
     position: 'absolute',
     width: borderWidth,
-    height: height,
+    height: height ?? '100%',
     top: 0,
     left: 0,
     backgroundColor: mainHexPallete.yellow[300],
-    zIndex: 0
+    zIndex: 1
   })
 };
--- a/app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
+++ b/app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
@@ -5,7 +5,8 @@ export const styles = {
     position: 'relative',
     width: width ?? '100%',
     height: height ?? '100%',
-    gridColumn: '1 / -1'
+    gridColumn: '1 / -1',
+    ...(height ? {} : { minHeight: 100 })
   }),
   border: (borderWidth: number, height?: number) => ({
     display: 'block',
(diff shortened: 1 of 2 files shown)


app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
```diff
@@ -1,20 +1,20 @@
 import { mainHexPallete } from '../theme/colors';
 
 export const styles = {
-  container: (width: number, height: number) => ({
+  container: (width?: number, height?: number) => ({
     position: 'relative',
-    width: width,
-    height: height,
+    width: width ?? '100%',
+    height: height ?? '100%',
```
kolibri753, line 7: If I'm not mistaken, in fill-mode the container might collapse unless it has a size. Maybe we add an aspect-ratio fallback when height isn't provided?
```ts
...(height == null ? { aspectRatio: '1 / 1' } : { height }),

```
or even like this
```ts
...(height ? { height } : { aspectRatio: ratio ?? '1 / 1' })
```

and let callers override the ratio  (ratio?: number | string;), or maybe it's too much

kolibri753 (approved):

app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
ruslansymonenko: You’re right, I didn’t consider that case. I’ll fix it now.

NatalyKrvch (approved):

app/shared/components/design-system/all-components/image-with-border/ImageWithBorder.styles.ts
ruslansymonenko: Thanks for catching that — you’re totally right, without a defined height the component just disappears.
Your version is a good approach, but in this particular case I think it doesn’t quite fit what I need.

The main goal here is for the component to fill its parent’s size.
I need to use this component later in my another task. I need to made image like on photo below. And I would like to set dimensions from parent component.
Знімок екрана 2025-11-06 о 18 12 53

When I use your version, **even with both width and height set on parent block**, the layout ends up looking different from what I expect. (maybe I did something wrong, i used ...(height == null ? { aspectRatio: '1 / 1' } : { height }), but as I understand, this behavior is supposed to happen with this setting)
Знімок екрана 2025-11-06 о 18 13 24
Знімок екрана 2025-11-06 о 18 13 42



So I think i can just add a minHeight.
```Javascript
    ...(height ? {} : { minHeight: 100 })
```
It’s a super simple fix that prevents the component from vanishing and works perfectly for both scenarios — when the height is set and when it’s not.
What do you think?
Знімок екрана 2025-11-06 о 18 13 24
Знімок екрана 2025-11-06 о 18 33 44

kolibri753: Sounds valid to me :)

stsvt (approved):

PR #887, opened 2025-11-10
refactor: Backend 

### LINKED ISSUE #815: (SP: 5)  Backend Refactor
## Description

There appears to be a few issues with backend as its written now. Those include:
1. It is impossible to skip through the definitions, you will always stop at the repository type and will not be able to find it itself
2. EACH Core Service completely duplicates the corresponding repository (fig.1 shows an example of how EACH of them is implemented), not to mention some of the Composed Services
3. The translation occurs on each service separately, although it can be done using one logic, which may not even be required (e.g. `ScietificWorksService`)
4. It does not matter to scope services because they are scoped in a container that is a singleton, therefore they have the same lifetime
4.1. Moreover, the container was designed to be unique per request
5. No type assistance after resolving services. In the example below artistryService is `any` so no type assistance from here on now
[In project](https://github.com/Liatoshynsky-Foundation/lf-client/blob/develop/app/api/compositions/data/route.ts#L27)
```
const artistryService = createRequestContainer().resolve('artistryService');
``` 
6. The container uses too verbose of a syntax and unnecessary compositions and decompositions that make the code harder to understand than it should be:
[In project](https://github.com/Liatoshynsky-Foundation/lf-client/blob/c004cbcddd41910a09d874fabc9dbcddca1c8960/app/di/modules/coreService.module.ts#L10)
```
foundationInfoService: asFunction(({ foundationInfoRepository }) =>
createFoundationInfoService(foundationInfoRepository)
).scoped()
```
with slight argument type change by wrapping repository in object can be shortened to
```
foundationInfoService: asFunction(createFoundationInfoService)
```
As `Awilix` provides out-of-the-box support for such syntax (refer to [Awilix docs](https://github.com/jeffijoe/awilix#injection-modes))
For Composed Services this goes to a new level, because dependencies are decomposed to be composed for the call:
[In project](https://github.com/Liatoshynsky-Foundation/lf-client/blob/c004cbcddd41910a09d874fabc9dbcddca1c8960/app/di/modules/composedServices.module.ts#L14)
```
headerService: asFunction(({ foundationInfoService, navigationService }) =>
createHeaderService({ foundationInfoService, navigationService })
).scoped(),
```
Repositories also have a certain interesting quirk:
[In project](https://github.com/Liatoshynsky-Foundation/lf-client/blob/c004cbcddd41910a09d874fabc9dbcddca1c8960/app/di/modules/repositories.module.ts#L10)
```
foundationInfoRepository: asFunction(() => foundationInfoRepository).scoped(),
```
This alone woudn't be a reason for change but in the light of previous issues, shall be addressed too.
All in all every entry in awilix can be changed to have a structure like this
```
<awilix registry entry>: asFunction(<constructor function name>).<lifetime>()
```

## Steps

1. Determine what is reasonable to change or add
2. Apply those
3. Check if everything is still working

## Expected Result

Nice and easy to use backend (if fortune smiles to us)

## Extra

Image

*Fig. 1*

yur4uwe:
## Description

Changes addressed:
- Removed core services to condense logic into composed services for general retrieving of values and repositories that query database directly
- Enhanced Zod parsing of pages by replicating structure of TipTap documents into zod schemas (fig. 1)
- Added simple npm script for static typechecks to reveal all hidden type errors
- Refactored Awilix container to create scopes for each request instead of single container, slightly modified container structure to resolve type blindness
- Next.js unfortunately doesn't allow a way to find a point in which all requests pass so we decided to drop single place for localization resorting to `LocalizeSchema`
- Now images use `generatedSrc` field because of quirk of zod transformations and structure of our code
- Replaced functions that manualy resolve localization

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

The changes made in this pr are simply cosmetical and do not change underlying logic, so to test it is to make sure everything works as had been up to this moment

## Video / Screenshots

*(Fig. 1)* New Structure of Zod Schemas for TipTap:
image

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---


app/lib/utils/translation/translate.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    expect(out).toEqual({ ambig: { en: 'OnlyEn' }, full: 'E' });
+  });
+
+  it('should not mutate the source object', () => {
+    const src = {
+      a: { en: 'X', uk: 'Y' },
+      b: [{ en: 'P', uk: 'Q' }]
+    };
+
+    // Sonnar may suggest structuredClone, but its a `window` method
```
NatalyKrvch, line 74: nice :)

NatalyKrvch (approved):
love your last commit message )
it seems nothing breaks, so approve

NatalyKrvch:
change pr name to match conventional commits, please

yur4uwe:
> change pr name to match conventional commits, please

I wnder if we have to do that for pr's? for commits, sure, but i don't recall conventions about pr

NatalyKrvch:
> > change pr name to match conventional commits, please
> 
> I wnder if we have to do that for pr's? for commits, sure, but i don't recall conventions about pr

image

we discussed it last sprint and now we have pull-request-workflow-guidelines.md in docs

yur4uwe:
> image
> we discussed it last sprint and now we have pull-request-workflow-guidelines.md in docs

Oh, i guess i was missing from that one

app/shared/components/blocks/privacy-policy/policy-section/PolicySection.test.tsx
```diff
@@ -3,7 +3,7 @@ import { render, screen } from '@testing-library/react';
 import React from 'react';
 
 import PolicySection from './PolicySection';
-import type { TipTapDoc } from '~/types/types/common.types';
+import type { TipTapDoc } from '~/types/types/tiptap.types';
```
kolibri753, line 7: Also there is a line in this test (`doc: import('~/types/types/common.types').TipTapDoc;`) that should be changed in the same way

app/types/page/about-us.types.ts
```diff
@@ -1,8 +1,9 @@
-import { TipTapDoc } from '~/types/types/common.types';
+import { TipTapDoc } from '~/types/types/tiptap.types';
 
 export interface IImageBlock {
   src: string;
   alt: string;
+  generatedSrc: string;
```
kolibri753, line 6: Since generatedSrc is being added in this PR, maybe better double-check that all the places creating IImageBlock are updated to provide it? Right now some mocks only set src/alt/caption.

app/infrastructure/repositories/pages-data/pagesData.repo.d.ts
```diff
@@ -0,0 +1,8 @@
+import { PageBaseMap } from '~/types/page/pagesBase.type';
+
+import { PageSlug } from '~/services/pages-data/schema-factory';
+
+export interface PagesDataRepository {
+  getBySlug: <S extends PageSlug>(slug: PageSlug) => Promise<PageBaseMap[S] | null>;
+  getDraftBySlug: <S extends PageSlug>(slug: PageSlug) => Promise<PageBaseMap[S] | null>;
```
kolibri753: I might be missing something, but as I see it, right now the generic S isn't really connected to the argument, since slug is typed as PageSlug. Maybe slug should use S instead, so calling getBySlug('about-us') returns the concrete page type instead of a union?

kolibri753 (approved):
Wow, nice job, I think it might be better to run `npm run typecheck` (thanks for adding this command to package.json, by the way) and double-check, because some of the errors seem to be introduced by this PR... but it could also be enough to create a small follow-up issue instead :)

alisa-korniienko:
We are not afraid, we are delighted with your attempts ☺  Great work!

alisa-korniienko (approved):

app/infrastructure/repositories/pages-data/pagesData.repo.d.ts
yur4uwe: those are type shenanigans of typesript, otherwise it would extend type to signature `Promise<AboutUsPage | PolicyPage | ResarchPage | null>`(names might be incorrect) and this type is incorrectly expanded at point of its application of pages which causes it to think that another page type is used in that file

app/types/page/about-us.types.ts
yur4uwe: Yeah i will double check this

app/shared/components/blocks/privacy-policy/policy-section/PolicySection.test.tsx
yur4uwe: Oh i should fix this strange syntax of inline imports

yur4uwe:
> we discussed it last sprint and now we have pull-request-workflow-guidelines.md in docs

fixed

yur4uwe:
> love your last commit message ) it seems nothing breaks, so approve

funniest thing was missing, I added `!` to make typescript ignore undefined type so that it won't annoy me

app/types/page/about-us.types.ts
yur4uwe: `IImageBlock|IQuoteBlock|IIntroSection|IFoundationInfo|IOurMission|ILiatoshynskyOffice|IFoundationFounders|IAboutUsPage`

I searched with this regex and it should have revealed every time `IImageBlock` is used so now i shouldn't have missed anything

app/infrastructure/repositories/pages-data/pagesData.repo.d.ts
yur4uwe: The type will exapand a little better to help intellisense pick up on it

PR #926, opened 2025-11-19
feat(ui): build archive case details page

### LINKED ISSUE #899: (SP: 3) Build “Case Details” page (list of documents)
## Description

Implement the Case Details page that displays detailed information about a specific archival case (справа) and the ordered list of documents, plus navigation back to the fund and prev/next case controls.

Scope & Requirements:

1. Header & Navigation
- "Повернутись" button returns to the parent Fund Details page
- Page title displays the current case name
2. Two-column layout
- Left column:
    - Шифр (index), Дати справи (date range), Аркушів (sheet count)
    - Primary action: "Переглянути PDF"
- Right column:
    - Ordered list of all documents in the case (titles, optional notes), numbered as per mockups
3. Sticky PDF button (designer note)
- The "Переглянути PDF" button must become sticky when the user scrolls down:
    - Sticks to top: 48px of the viewport
    - On scroll up, it stays fixed until it reaches its original position, then returns to normal flow
- The button remains visible while the user reads the documents list
4. Prev/Next case navigation (designer note)
- Controls: "Попередня справа", "Наступна справа" with dynamic labels (case number + title)
- This navigation block is located at the bottom of the page
- Order follows the sequence shown in the Fund’s cases table
- Edge cases:
    - First case: no "Попередня справа"
    - Last case: no "Наступна справа"
- Labels update to reflect the adjacent case titles
5. Responsive behavior:
- For screen widths ≥ 768px the layout consists of two columns:
    - left column – case details (code/identifier, date range, page count, etc.)
    - right column – ordered list of documents
- For screen widths < 768px the layout switches to a single-column (stacked) layout:
    - case details block goes first
    - documents list goes below it
- Spacing and typography should follow the Figma design for each breakpoint

Expected Result:
Case title and case details render correctly
PDF button opens the correct file and behaves as sticky per spec
Ordered documents list is complete and numbered
Back to fund works
Prev/Next load the correct adjacent cases; hidden on first/last as specified
Layout and states match Figma across breakpoints

## Acceptance criteria

1. Case details(index, dates, sheets) displayed
2. "Переглянути PDF" is sticky with top: 48px, returns to normal on scroll up
3. Ordered list of documents rendered and accessible
4. "Повернутись" navigates back to Fund details page
5. Prev/Next buttons follow fund order, hidden appropriately, labels update dynamically
6. Matches Figma on desktop/tablet/mobile

## Extra

Image

kolibri753:
## Description

Add UI layer for the Case Details page in the Archive section.
The page shows a single archival case, its ordered documents list, a sticky “View PDF” action, and navigation to the previous/next case.

At this stage the page uses a mocked case (MOCK_ARCHIVE_CASE) and does not yet load data based on real fund / case IDs. Changing the slug only updates the URL, not the content. Wiring to real data is planned for a follow-up PR.

The page is available at /[lang]/archive/[fund]/[case]. It reads lang and fund from params, configures next-intl with lang, and builds the “Back” URL to the Fund Details page from lang + fund. The case itself is always taken from the mock and mapped into a simple { id, title } document shape for the UI.

The main layout is handled by ArchiveCaseDetails, which composes:

- a BackLink that uses the shared CustomLink + Svg stack and navigates back to the parent fund;
- a Meta column with reference code, date range, optional sheets count and the primary “View PDF” action;
- a Documents column with the ordered list of documents;
- a Navigation block that shows “previous” and/or “next” case controls when provided.

The "View PDF" button is implemented as a reusable PdfButton component (MUI Button acting as an external link).
Navigation at the bottom shows previous/next case buttons when data is provided. Each side is a NavItem with an arrow icon, a link to caseLink.href, and the neighbour’s index + title; with the mock both neighbours are present, so both buttons work but still show the same mocked case.

All labels come from a new archiveCase i18n namespace and are passed into ArchiveCaseDetails via a single labels prop, so the tree stays free of hardcoded strings.

## Type of change

- [x] New feature

## How to test

1. Visit /uk/archive/2/op1-spr3 and /en/archive/2/op1-spr3. Confirm the back link, case title, meta, documents list, and navigation all render in correct language.
2. Check the "sticky" PDF button - resize the window to a desktop/tablet width:
- scroll down: the PDF button in the left column should stick and stay visible while you move through the documents;
- scroll back up: it should "release" and return to its original place.
3. Check bottom navigation (+ edge cases):
- both Prev and Next controls should be visible for the mock;
- they should have correct labels and show the mock index + title below each button;
- manually make one of them null in archiveCase.mock.ts (prev: null) and check the layout.

P.S. It would be best to check the responsiveness to ensure it matches the Figma design.

## Video / Screenshots

image

https://github.com/user-attachments/assets/7cced3cb-c53d-4c25-80ad-e6dbbd267875

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---


app/shared/components/blocks/archive-case-details/ArchiveCaseDetails.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+      md: '120%'
+    },
+    textTransform: 'uppercase',
+    mb: {
+      sm: '8px',
+      md: '32px'
+    },
+    gridColumn: {
+      xs: '1 / -1',
+      sm: '1 / -1'
```
yur4uwe: Unnecessary repetition of gridColumn, it doesn't change with screen width

app/shared/components/blocks/archive-case-details/ArchiveCaseDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export interface ArchiveCaseDetailsProps {
+  title: string;
+  index: string;
+  dateRange: string;
+  sheetsCount?: number;
+  pdfUrl: string;
+  documents: ArchiveCaseDocument[];
+  fundHref: string;
+  prevCase?: ArchiveAdjacentCase | null;
```
yur4uwe: Why do you need `ArchiveAdjacentCase | null | undefined`, you can make null check for undefined

app/shared/components/blocks/archive-case-details/meta/Meta.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <Typography sx={styles.metaLabel}>{labels.code}</Typography>
+        <Typography sx={styles.metaValue}>{index}</Typography>
+      </Box>
+
+      <Box sx={styles.metaBlock}>
+        <Typography sx={styles.metaLabel}>{labels.dates}</Typography>
+        <Typography sx={styles.metaValue}>{dateRange}</Typography>
+      </Box>
+
+      {typeof sheetsCount === 'number' && (
```
yur4uwe: Why do you make type check here instead of `undefined` check

app/shared/components/blocks/archive-case-details/pdf-button/PdfButton.tsx
yur4uwe, line 1: I think you could've extended our custom button component to accept some kind of flag to use external link

kolibri753: I actually thought about extending our custom button, but because of the design I have to render this same button separately for mobile and desktop. To avoid duplicating the same properties in a few places (and on another archive page as well), it felt safer to wrap it in a small PdfButton. This way if the button changes later, we only update it in one place

yur4uwe: i probably missed this in my first time viewing, but why do you separately render `view PDF` on big screens in `Meta` and  inside `Documents` on small ones, i simply can't see why you decided to do that

kolibri753: On large screens, the design places the PDF button in the meta column and keeps it sticky. On small screens, we only have one column, so if the button remained in the meta section, it would scroll out of view as the user reads the list.

To keep it visible while the documents list is on screen, I rendered the same PdfButton twice and toggled them using breakpoints.

yur4uwe: Then wouldn't it be better to ditch columns and render it with left margin for different screens width, this will keep it in one place without need to place it inside another block?

yur4uwe: But why would it scroll out of view? Did you set overflow on `Meta` to clip?

kolibri753: The issue is that the sticky button lives inside the Meta section, and `position: sticky` is bounded by that section's box.
Meta sits above the long documents list, so once the bottom of Meta scrolls past, a sticky button inside it will also scroll away while the user is still reading the documents.
That's why on mobile I attach a sticky button to the documents block instead: it keeps the same PdfButton visible for the whole scroll range of the list.

P.s. I did think about a single button with margins, but for me the layout has two clear logical areas, so i decided to stick with them :)

yur4uwe: Then wouldn't making 3 blocks inside archive page solve this issue by introducing pdf button as container in the middle, bound by archive page block and effectively solving the issue of having 2 different buttons

kolibri753: I see your point, and technically I could add a third block in the middle with a single sticky button. If I understood you correctly, the main goal is to avoid having a dedicated PdfButton component and the three lines `<Box><PdfButton /></Box>` "duplication".

I believe we will reuse PdfButton on another archive page, so the component itself still brings value. Refactoring to a central PDF block would add more grid/order complexity for not much gain, as I see it, so I don't feel this trade-off is needed right now. But I can try to implement this to see the difference more clearly)

yur4uwe: Well, even if we don't extend button, having a box jump containers is more unreliable than having additional container inside page and solve its positional problems, i would argue

kolibri753: Okay, okay... I'll look more into your approach :)

ruslansymonenko (approved):

yur4uwe (approved):

NatalyKrvch (approved):

PR #1038, opened 2025-12-08
feat: migrate hardcoded UI content from biography page to database

### LINKED ISSUE #1029: (SP: 2) Migrate hardcoded UI content from Biography page to database
## Description

The Biography page currently contains content that is hardcoded on the frontend. The goal is to move this content to the database and update the app so the page fetches the content via API and renders it dynamically.

## Steps

1. Identify all hardcoded Biography page content (titles, texts, quotes, images, lists, etc.).
2. Create the DB structure for Biography content.
3. Populate the database with the current Biography content.
4. Implement or update an API endpoint to retrieve Biography content.
5. Update the frontend to remove hardcoded data and fetch content from the API (including proper typing and 
loading/error handling as per project conventions).
7. Verify the UI matches the current design/behavior after migration (no regressions).

## Acceptance criteria

1. No hardcoded content remains on the Biography page.
2. Biography content is loaded from the database via API and rendered correctly.
3. Data structure fully covers all Biography page blocks and is properly validated by Zod.
4. UI stays identical to the previous implementation (layout/content order/visuals).

Taras-ep:
## Description

The Biography page contained content that was hardcoded on the frontend. The goal was to move this content to the database and update the app so that the page fetches the content via API and renders it dynamically.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. npm run dev
2. switch to /biography
3. make sure that all the content renders according to the Figma design

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues
- [x] Documentation updated (if applicable)
- [x] All comments and requested changes addressed

---


app/[lang]/biography/BiographyContent/BiographyContent.tsx
```diff
@@ -11,16 +10,14 @@ import type {
...
 
 import ContentBlock from '~/shared/components/design-system/all-components/content-block/ContentBlock';
 import ExcerptBlock from '~/shared/components/excerpt-block/ExcerptBlock';
 import ImageWithCaption from '~/shared/components/image-with-caption/ImageWithCaption';
 import YearWithLine from '~/shared/components/year-with-line/YearWithLine';
 
 export function BiographyContent({ data }: Readonly<BiographyContentProps>) {
-  const locale = useLocale();
-
   function renderChrolologyList(item: ChronologyList, key: string): ReactNode {
```
Comment (kolibri753) [inline app/[lang]/biography/BiographyContent/BiographyContent.tsx]: Since you've already touched this file, can you also fix this typo: `renderChrolologyList` -> `renderChronologyList`? Thanks in advance :)

app/[lang]/biography/page.tsx
```diff
@@ -1,26 +1,45 @@
...
+  const { lang } = await params;
+  setRequestLocale(lang);
+
+  const pageService = await createRequestContainer().resolve('pagesDataService');
+
+  const page = await pageService.getPageData('biography', lang);
+
+  if (!page) {
+    return <></>;
+  }
```
Comment (kolibri753) [inline app/[lang]/biography/page.tsx]: What about returning notFound() or CustomNotFoundPage?

app/[lang]/biography/page.tsx
```diff
@@ -1,26 +1,45 @@
...
+
   if (isProductionMode()) {
     return <UnderDevelopment />;
   }
```
Comment (kolibri753) [inline app/[lang]/biography/page.tsx]: would it make sense to early-return on isProductionMode() **before** resolving the container and calling getPageData, so we don't hit the DB when the page is hidden

Comment (Taras-ep) [reply app/[lang]/biography/page.tsx]: Yep, good point. We can early-return to avoid resolving the container and hitting the DB when the page is hidden. I'll update the flow.

app/[lang]/biography/BiographyContent/BiographyContent.tsx
Comment (Taras-ep) [reply app/[lang]/biography/BiographyContent/BiographyContent.tsx]: Good point! Updated

app/[lang]/biography/page.tsx
Comment (Taras-ep) [reply app/[lang]/biography/page.tsx]: Good idea! Implemented.

app/validators/pagesSchemas/pages/biography.schema.ts
```diff
@@ -0,0 +1,100 @@
...
+    generatedSrc: `/api/blob-url?folderName=photos&blobName=${image.src}`
+  }));
+
+const HeroSectionBlockSchema = z.object({
+  title: translatedFieldSchema,
+  image: HeroImageSchema,
+  quote: QuoteSchema,
+  biographyText: LocalizedTipTapDocSchema,
+  noteText: translatedFieldSchema
+});
```
kolibri753, line 83: I might be missing some context, but is title here actually used? HeroSection still takes its title from translations - should we either wire this CMS title into the UI or drop it to avoid a "dead" field?

app/types/page/biography.types.ts
```diff
@@ -65,27 +67,22 @@ export interface BiographyContentProps {
...
   caption: HeroImageCaption;
 }
 export interface BiographyHeroData {
   quote: BaseQuoteBlock;
   image: HeroImage;
-  biographyText: LocalizedTipTapDoc;
-  noteText: LocalizedTipTapDoc;
+  biographyText: TipTapDoc;
+  noteText: string;
 }
```
kolibri753: Since we already have BiographyPage from the Zod schema, would it make sense to type this as BiographyPage['blocks']['heroSection'] so it always matches the page schema, or do we prefer keeping a separate UI-only shape here?

app/types/page/biography.types.ts
```diff
@@ -56,7 +58,7 @@ export interface FullWidthImage extends BiographyContentItemBaseProps<ContentTyp
 export type BiographyContentItem = ChronologyList | ExcerptBlockItem | OnlyImageBlock | FullWidthImage;
 
 export interface BiographyContentBlock {
-  yearTitle: string | undefined;
+  yearTitle: string | null;
   items: BiographyContentItem[];
 }
```
kolibri753, line 63: Similar question here: could this be BiographyPage['blocks']['biographyContent'][number] to stay in sync with the schema, or is there a reason to keep this interface decoupled from the page schema?

kolibri753 (commented):
The comments about reusing BiographyPage['blocks'] types are non-blocking suggestions. If you'd prefer to keep it as is and maybe tweak types in a follow-up, I'm OK with that - just ping me :)

app/validators/pagesSchemas/pages/biography.schema.ts
Taras-ep: Good catch about the title on HeroSection - it isn’t used in the UI, we still take the title from translations. I removed this field from the schema and DB to avoid keeping a dead/unused CMS field.

app/types/page/biography.types.ts
Taras-ep: Good point, thanks!
I’ve renamed BiographyHeroData to HeroSectionBlock to match the HeroSectionBlockSchema.
I’d prefer to keep this as a UI-level type rather than BiographyPage['blocks']['heroSection']. getPageData('biography', lang) already returns data resolved for the current locale (biographyText: TipTapDoc, strings instead of translated field objects), so the runtime shape is slightly different from the raw page schema, even though the field names are now in sync with it.

app/types/page/biography.types.ts
Taras-ep: Good point, thanks!
I’ve aligned the naming here as well (BiographyContentBlock -> BiographyContentBlockSchema)

app/types/page/biography.types.ts
```diff
@@ -19,21 +16,26 @@ export enum ImagesSizes {
...
 
 interface ChrolologyListItem {
```
kolibri753: the typo is still present :)

kolibri753 (approved):

alisa-korniienko (approved):

app/[lang]/biography/BiographyContent/BiographyContent.test.tsx
```diff
@@ -43,116 +43,115 @@ jest.mock('~/shared/components/year-with-line/YearWithLine', () => {
   };
 });
 
+const t = (uk: string, _en: string) => uk;
+
 const makeTipTapDoc = (text: string): TipTapDoc =>
   ({
+    type: 'doc',
```
Comment (MaksFullJs) [inline app/[lang]/biography/BiographyContent/BiographyContent.test.tsx]: It is not essential but you are not creating real tiptap item when you write `type: 'doc'`. You will have just a string. You should use TipTapNodeTypes and choose what type you need doc or paragraph or text. In that case you do not have to write `as TipTapDoc` in the bottom.

MaksFullJs (approved):

app/types/page/biography.types.ts
Taras-ep: Thanks! Fixed.

app/[lang]/biography/BiographyContent/BiographyContent.test.tsx
Comment (Taras-ep) [reply app/[lang]/biography/BiographyContent/BiographyContent.test.tsx]: Thanks! Updated. I've replaced string literals with TipTapNodeTypes and removed the explicit cast. The helper now returns a fully typed TipTapDoc.

PR #1047, opened 2025-12-09
Fix: console errors in useTableFilters hook and Search component

### LINKED ISSUE #1046: (SP: 0.5) Search & Filters Rendering Errors
## Summary

Several errors occurred when using the search autocomplete and table filter functionality.
These issues produced runtime warningsdue to incorrect DOM structure and side-effects triggered during render.

## Severity

- [ ] Blocker
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

## Steps to reproduce

1. Open any page containing the Search autocomplete component.
2. Type into the search input.
3. Open the dropdown list of options.
4. Navigate away and back (to trigger hydration).
5. Change search params through the filters (triggers useTableFilters).
6. See the console

## Logs / Console output

useTableFilters.ts:41 Cannot update a component (Router) while rendering a different component (WorkTableSection). To locate the bad setState() call inside WorkTableSection, follow the stack trace as described in https://react.dev/link/setstate-in-render

Search.tsx:108 A props object containing a "key" prop is being spread into JSX: let props = {key: someKey, tabIndex: ..., role: ..., id: ..., onMouseMove: ..., onClick: ..., onTouchStart: ..., data-option-index: ..., aria-disabled: ..., aria-selected: ..., className: ..., children: ...}; <li {...props} /> React keys must be passed directly to JSX without using spread:

Search.tsx:109 In HTML,` <li> `cannot be a descendant of `<li>`. This will cause a hydration error.

Search.tsx:108 `<li>` cannot contain a nested `<li>`. See this log for the ancestor stack trace.

## Acceptance criteria

- [ ] No console warnings during search or navigation

qqwz0:
## Description

This PR fixes several issues related to the Search autocomplete component and table filters.

1. Fixed useTableFilters side-effect during render
2. Prevented key from leaking through spread in Search renderOption
3. Fixed hydration error caused by nested <li> elements
4. Fixed virtualization layout collapse

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Open any page that uses the Search component.
2. Type into the search input to open autocomplete.
3. Verify that no console warnings appear

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---


NatalyKrvch (approved):

Taras-ep (approved):

app/shared/hooks/use-table-filters/useTableFilters.ts
```diff
@@ -46,25 +45,23 @@ export function useTableFilters<P extends TableParams>(initialParams: P) {
...
-        return next;
-      });
+      const next = { ...params, [key]: value };
+
+      setParams(next);
+      syncUrl(next);
     },
-    [syncUrl]
+    [params, syncUrl]
   );
```
kolibri753: I'm a bit worried about using params from the closure here.
In React, state updates can be batched, and when we build the next value from a captured variable instead of prev, we might accidentally use an older version of the state if setParam is called a few times in a row.
React docs usually recommend the "functional" form when the next state depends on the previous one.
What do you think about something like:
```ts
setParams(prev => {
  const next = { ...prev, [key]: value };
  syncUrl(next);
  return next;
});
```
This way we always work with the latest state React gives us, which should be a bit safer for filters

app/shared/hooks/use-table-filters/useTableFilters.ts
```diff
@@ -46,25 +45,23 @@ export function useTableFilters<P extends TableParams>(initialParams: P) {
...
-          return next;
-        });
+        const next = { ...params, [key]: value };
+
+        setParams(next);
+        syncUrl(next);
       }, 400),
-    [syncUrl]
+    [params, syncUrl]
   );
```
kolibri753, line 75: Same concern as in setParam: we rely on params from the closure. Maybe reuse the setParams(prev => { ... }) pattern here and then drop params from deps, so debounce works with the latest state

app/shared/components/search/Search.tsx
```diff
@@ -103,28 +103,37 @@ export const Search = <T extends { title?: string | { en?: string; uk?: string }
...
+          variant="customMedium16"
+          sx={{
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+            display: '-webkit-box',
+            WebkitLineClamp: 2,
+            WebkitBoxOrient: 'vertical',
+            lineHeight: '1.2'
+          }}
+        >
```
kolibri753, line 132: we used to have wordBreak: 'break-word' here, not sure, but might be worth keeping it?

app/shared/hooks/use-table-filters/useTableFilters.ts
qqwz0: you're absolutely right, and under normal circumstances your solution would be the safest approach.

however, in this case` syncUrl()` internally calls `router.replace()`. when this runs inside the updater function, it becomes a render-phase side effect, which causes the Next.js App Router error I mentioned in the issue. Next.js does not allow navigation actions during render, even if the state update itself is correct.

and to avoid both the stale closure issue and the Router error, syncUrl(next) needs to run after the state update, not inside the updater. 

so keeping `syncUrl(next)` outside the updater is required for compatibility with the App Router, while still ensuring the state is always up to date.

kolibri753: Got it, thanks for the explanation, that makes sense. Could this be an alternative: keep setParams(prev => ({ ...prev, [key]: value })) and move syncUrl(params) into a useEffect on params, so the updater stays pure and we avoid the closure issue? But it's up to you - if you’re confident the current usage pattern is safe enough for this hook, I'm okay with it :)

bandvov: @qqwz0 try to use `shallow` flag for router. Maybe this will help avoid rerender

app/shared/components/search/Search.tsx
qqwz0: done!

app/shared/hooks/use-table-filters/useTableFilters.ts
qqwz0: thanks for the suggestion!

sadly, the App Router doesn’t support the `shallow` option on `router.replace`.

I switched to the variant that @kolibri753 suggested (moved `syncUrl(params)` into a `useEffect` on `params`, so navigation happens after render and the updater stays pure). thanks, you helped me a lot!

that should avoid both the stale-closure risk and the error.

kolibri753 (approved):

PR #1103, opened 2025-12-19
feat: delete hardcoded paddings and add margins according to design

### LINKED ISSUE #1077: (SP: 0.5) Move EnhancedTable padding control to parent components
## Description

Currently, the `EnhancedTable` component has a hardcoded vertical padding (`py: 20`) defined in its internal `enhancedTableStyles`. This limits the component's flexibility, as the spacing cannot be adjusted based on the context where the table is used.

We need to remove this internal styling and update the component to accept an `sx` prop. This change requires finding all current usages of `EnhancedTable` and manually re-applying the padding to ensure no visual regressions occur on existing pages.

## Steps

1.  **Update Component Logic:**
    * Modify `EnhancedTableProps` to accept an optional `sx` prop (type `SxProps<Theme>`).
    * Update the root `Box` in `EnhancedTable.tsx` to merge the internal `styles.root` with the passed `sx` prop.
2.  **Update Styles:**
    * Remove `py: 20` from the `root` object in `EnhancedTable.styles.ts`.
3.  **Refactor Usages:**
    * Search the codebase for all instances of `<EnhancedTable />`.
    * For each instance, pass the original padding (equivalent to `py: 20` or specific design requirements) via the new `sx` prop or a wrapper container to maintain the current visual appearance.

## Acceptance criteria

1.  The `EnhancedTable` component does not have default vertical padding (`py: 20`) in its base styles.
2.  The component accepts and applies custom styles via the `sx` prop.
3.  All existing pages using the table look identical to how they looked before the change (no visual regressions).

## Expected Result

The `EnhancedTable` component becomes more reusable and composable, allowing parent components to dictate the spacing and layout around the table.

MaksFullJs:
## Description

Removed hardcoded padding styles in EnhancedTable component and added `tableContainerSx` prop to allow managing spacing externally. And used this prop to define margins according to design

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Just visit any page with table(Artistry, Research and Academic work, Fund Detail) and check margins

## Checklist

- [ ] PR name follows the naming convention
- [ ] Tests pass successfully
- [ ] No Sonar issues

---

### DIFF
--- a/app/shared/components/enhanced-table/EnhancedTable.tsx
+++ b/app/shared/components/enhanced-table/EnhancedTable.tsx
@@ -1,6 +1,15 @@
 'use client';
 
-import { Box, CircularProgress, Paper, SxProps, Table, TableBody, TableContainer, Theme } from '@mui/material';
+import {
+  Box,
+  CircularProgress,
+  Paper,
+  type SxProps,
+  Table,
+  TableBody,
+  TableContainer,
+  type Theme
+} from '@mui/material';
 import {
   ColumnDef,
   ColumnFiltersState,
--- a/app/shared/components/enhanced-table/EnhancedTable.tsx
+++ b/app/shared/components/enhanced-table/EnhancedTable.tsx
@@ -1,7 +1,6 @@
 'use client';
 
-import { Theme } from '@emotion/react';
-import { Box, CircularProgress, Paper, SxProps, Table, TableBody, TableContainer } from '@mui/material';
+import { Box, CircularProgress, Paper, SxProps, Table, TableBody, TableContainer, Theme } from '@mui/material';
 import {
   ColumnDef,
   ColumnFiltersState,
--- a/app/shared/components/enhanced-table/EnhancedTable.tsx
+++ b/app/shared/components/enhanced-table/EnhancedTable.tsx
@@ -1,6 +1,7 @@
 'use client';
 
-import { Box, CircularProgress, Paper, Table, TableBody, TableContainer } from '@mui/material';
+import { Theme } from '@emotion/react';
+import { Box, CircularProgress, Paper, SxProps, Table, TableBody, TableContainer } from '@mui/material';
 import {
   ColumnDef,
   ColumnFiltersState,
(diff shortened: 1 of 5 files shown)


app/shared/components/enhanced-table/EnhancedTable.tsx
```diff
@@ -1,6 +1,7 @@
 'use client';
 
-import { Box, CircularProgress, Paper, Table, TableBody, TableContainer } from '@mui/material';
+import { Theme } from '@emotion/react';
```
NatalyKrvch: import **type** { Theme } from '@emotion/react';
because you use Theme only for typing

NatalyKrvch (approved):

app/shared/components/enhanced-table/EnhancedTable.tsx
MaksFullJs: yes, thanks, also noticed this, fix is coming

DenisGordProgrammer (approved):

kolibri753 (approved):

app/shared/components/enhanced-table/EnhancedTable.tsx
MaksFullJs: > import **type** { Theme } from '@emotion/react'; because you use Theme only for typing

should be fixed

NatalyKrvch: almost there :)

it looks like Theme and SxProps are still imported as values. since they are purely ts types, it's best practice to add the type keyword explicitly (import { type Theme, ... }). this helps the bundler understand that these are just types and can be safely stripped out of the final js bundle

MaksFullJs: > almost there :)
> 
> it looks like Theme and SxProps are still imported as values. since they are purely ts types, it's best practice to add the type keyword explicitly (import { type Theme, ... }). this helps the bundler understand that these are just types and can be safely stripped out of the final js bundle

Fixed

PR #1354, opened 2026-05-26
feat(intro): implement one-time intro animation via session storage

dest411:
## Description

Implemented a feature to ensure the intro animation on the homepage is only played once per session, improving the overall user experience. This was achieved by creating a new IntroAnimationContext that tracks the user's state using sessionStorage. The IntroAnimation component now checks this context before rendering.

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Open the homepage (/ or /[lang]).
2. Verify: The intro animation should play fully.
3. Navigate to any other page (e.g., About Us) and return to the homepage.
4. Verify: The homepage content should load instantly without the animation.
5. Reload the page (F5).
6. Verify: The animation is skipped, and content loads instantly.
7. Close the browser tab/window completely, then open a new tab and go to the homepage.
8. Verify: The intro animation should play again (since the session was cleared).

## Video / Screenshots

https://github.com/user-attachments/assets/728df7a6-a86d-4225-8260-db7a39a474ad

## Expected result:
The intro animation plays only on the very first visit to the homepage during a browser session. Subsequent navigations or reloads within the same session bypass the animation, allowing users immediate access to the content.

Closes: https://github.com/Liatoshynsky-Foundation/lf-client/issues/1325

### DIFF
--- a/app/shared/components/blocks/home-page-hero/animation/IntroAnimation.tsx
+++ b/app/shared/components/blocks/home-page-hero/animation/IntroAnimation.tsx
@@ -6,15 +6,39 @@ import { FC, ReactNode, useState } from 'react';
 import { wordEightPaths } from './logoSVGPaths';
 import { WordMorpher } from './WordMorpher';
 
+import { useIntroAnimation } from '~/shared/context/IntroAnimationContext';
+
 interface IntroAnimationProps {
   children: ReactNode;
   testID?: string;
 }
 
 export const IntroAnimation: FC<IntroAnimationProps> = ({ children, testID = 'intro-animation' }) => {
+  const { hasSeenIntro, markIntroAsSeen, isInitialized } = useIntroAnimation();
+
   const [showIntro, setShowIntro] = useState(true);
   const [showExpansion, setShowExpansion] = useState(false);
   const [showContent, setShowContent] = useState(false);
+
+  if (!isInitialized) {
+    return (
+      <div
+        style={{
+          position: 'fixed',
+          top: 0,
+          left: 0,
+          width: '100vw',
+          height: '100vh',
+          backgroundColor: '#ffffff',
+          zIndex: 9999
+        }}
+      />
+    );
+  }
+
+  if (hasSeenIntro) {
+    return <>{children}</>;
+  }
 
   const expansionVariants = {
     initial: {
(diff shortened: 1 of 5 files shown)


app/shared/components/blocks/home-page-hero/animation/IntroAnimation.tsx
```diff
@@ -6,16 +6,40 @@ import { FC, ReactNode, useState } from 'react';
...
+    return (
+      <div
+        style={{
+          position: 'fixed',
+          top: 0,
+          left: 0,
+          width: '100vw',
+          height: '100vh',
+          backgroundColor: '#ffffff',
+          zIndex: 9999
```
DenisGordProgrammer, line 33: Didn't we found any other approaches? That line looks soooo weird, and could cause so much problems in future.

next.config.ts
```diff
@@ -20,6 +20,12 @@ const nextConfig: NextConfig = {
         hostname: 'images.unsplash.com',
         port: '',
         pathname: '/**'
+      },
+      {
+        protocol: 'https',
+        hostname: 'pub-2b50c59c64954ab89b7837f9f4607e12.r2.dev',
```
DenisGordProgrammer: And i want double check that this is not sensitive info

DenisGordProgrammer (changes requested):

next.config.ts
dest411: You're right

app/shared/components/blocks/home-page-hero/animation/IntroAnimation.tsx
dest411: I was just sticking to the existing template to keep this PR focused. Refactoring the style here is beyond the scope, I think it's worth creating a new ticket for this

DenisGordProgrammer (approved):
okay, BUT, we should not forget about TODO: Refactor z-index to use a shared theme constant

navimov (approved):
Great work. I think we also need to create a task to replace the directly written color codes with constants from the color file.

PR #1376, opened 2026-06-02
refactor(theme): convert hardcoded z-indices to design system constants

Yushchyk-Roman:
## Description
- Implement theme variables for z index
- refactor and convert old hardcoded values into new constant variables.

## Type of change
- [x] Refactoring / Tech debt

## How to test
So to test it you need to run client side locally and check:
- if main animation is working correctly.
- if navbar is above all content
- if main logo and support button is above mobile overlay.
- if years navigation is above all content in biography page.

## Video / Screenshots

https://github.com/user-attachments/assets/33173ee4-98ee-4e2e-b85b-b5c735df563e

Vid. 1. ensure that mobile menu is working correctly

https://github.com/user-attachments/assets/19b6698d-5efb-422d-b6ab-96e7e9f8b236

Vid. 2. ensure that biography navigation bar is working

https://github.com/user-attachments/assets/47c0f7e0-a935-4e46-9a42-3883c7276dab

Vid. 3. Check if main animation is above all content

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues
- [x] All comments and requested changes addressed

## Acceptance Criteria:

- [x] Hardcoded z-index values are removed from all components.
- [x] Stacking order is managed via the centralized theme definition.
- [x] No regressions in UI layering (modals, tooltips, and headers behave as expected).

---

closes #1355

### DIFF
--- a/app/shared/components/design-system/all-components/theme/Theme.ts
+++ b/app/shared/components/design-system/all-components/theme/Theme.ts
@@ -172,6 +172,7 @@ declare module '@mui/material/styles' {
     introAnimationBackground: number;
     introAnimationExpansion: number;
     mobileOverlay: number;
+    headerAppBar: number;
     mobileNavButton: number;
     modalCloseButton: number;
     stickyYearsTab: number;
@@ -221,7 +222,7 @@ export const theme = createTheme({
     introAnimationBackground: 3000,
     introAnimationExpansion: 2999,
     mobileOverlay: 1050,
-    appBar: 1100,
+    headerAppBar: 1100,
     mobileNavButton: 1100,
     modalCloseButton: 100,
     stickyYearsTab: 50,
--- a/app/shared/components/design-system/all-components/theme/Theme.ts
+++ b/app/shared/components/design-system/all-components/theme/Theme.ts
@@ -166,6 +166,18 @@ declare module '@mui/material/Button' {
     tertiary: true;
   }
 }
+
+declare module '@mui/material/styles' {
+  interface ZIndex {
+    introAnimationBackground: number;
+    introAnimationExpansion: number;
+    mobileOverlay: number;
+    mobileNavButton: number;
+    modalCloseButton: number;
+    stickyYearsTab: number;
+    cursor: number;
+  }
+}
 export const theme = createTheme({
   palette: {
     primary: {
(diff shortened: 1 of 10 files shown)


DenisGordProgrammer (changes requested):
Nice, thats what we needed, but, general feedback: 
1.introAnimationBackground and mobileNavButton had just the same z-index.If we will NEVER see both elements in one page, thats okay, but, if not?
2.cursor: 9999 its pretty big number, different libraries uses it for some widgets or analytics.My suggestion is hold top lvls of z-index a bit lower, like, >9000, to reduce impact of overlays of other services
3.i couldnt find .appBar that you mension in headers styles, in theme.ts, so i think you use standart MUI (its 1100), but, in that case our mobileOverlay(1099) positions right under appBar. Double-check the behaivor in that case: Make sure that when the mobile menu is opened, the header (appBar) actually overlaps the overlay, or vice versa - that the zIndex logic does not break standard MUI components.

app/shared/components/design-system/all-components/theme/Theme.ts
```diff
@@ -220,11 +220,12 @@ export const theme = createTheme({
   zIndex: {
     introAnimationBackground: 3000,
     introAnimationExpansion: 2999,
-    mobileOverlay: 1099,
-    mobileNavButton: 3000,
+    mobileOverlay: 1050,
+    appBar: 1100,
```
DenisGordProgrammer: nice, but, actually it could be controversal, and become a conflict, as you use same reserved word. i suggest you to use some unique, like, idk, appNavBar or mainAppBar, use which more correct to the situation

DenisGordProgrammer (changes requested):

app/shared/components/design-system/all-components/theme/Theme.ts
Yushchyk-Roman: Thanks for the good advice. I initially thought the default name would be fine, but I totally agree that using a unique name is a much safer approach to avoid future conflicts. I've renamed it to headerAppBar. Thanks for the clear review!

DenisGordProgrammer (approved):
nice

vladashvch (approved):
LGTM

